PR #2, opened 2022-05-05
Single message echo server

jonmanahan:
Once the server is spun up via building and running the application in IntelliJ, a curl request can be sent to the corresponding port (4424) establishing a connection with the server, and one single message can be entered into the terminal to be echoed back before the server and application close.


elijahsullivan (commented):
How do I run the application locally? Could you include information in the README?

jonmanahan:
> How do I run the application locally? Could you include information in the README?

I meant to do this before, apologies, I can't believe I missed this

README.md
```diff
@@ -1,4 +1,17 @@
 # waiter
 
 ## Description
-A server that establishes a connection with the client, recieves their message and echos it back to them
\ No newline at end of file
+A server that establishes a connection with the client, receives their message and echos it back to them
+
+## Prerequisites
+- gradle build
```
elijahsullivan, line 7: Does this assume I already have gradle installed?

jonmanahan: That is a good point, It does, I should put that in the prereqs to install it, woops 😮

elijahsullivan: Does this also assume I have Java installed? If so, which version of Java?

jonmanahan: Another good point, I missed them all! 😵‍💫 The version is Java 18.0.1.1

.idea/gradle.xml
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
+  <component name="GradleMigrationSettings" migrationVersion="1" />
```
elijahsullivan, line 3: I see this `.idea` directory. Does this directory need to be tracked in source control? Do I need to be using a JetBrains IDE to work with this project?

app/src/main/java/waiter/Listener.java
```diff
@@ -0,0 +1,15 @@
...
+import java.io.IOException;
+import java.net.ServerSocket;
+
+public class Listener implements Awaiter {
+    ServerSocket server;
+
+    Listener(ServerSocket server) {
+        this.server = server;
+    }
+    public Connection awaitClient() throws IOException {
```
elijahsullivan, line 12:
```suggestion
    }
    
    public Connection awaitClient() throws IOException {
```

app/src/main/java/waiter/ClientConnection.java
```diff
@@ -0,0 +1,27 @@
...
+import java.net.Socket;
+
+public class ClientConnection implements Connection {
+
+    private final Socket client;
+
+    ClientConnection(Socket client) {
+        this.client = client;
+    }
+    public Message read() throws IOException {
```
elijahsullivan, line 16:
```suggestion
    }
    
    public Message read() throws IOException {
```

app/src/test/java/waiter/CommunicatorTest.java
```diff
@@ -0,0 +1,21 @@
...
+
+import java.io.IOException;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class CommunicatorTest {
+
+    @Test
+    void establishesCommunication() throws IOException {
+        String userInput = "Doesn't matter whats here, simulating any user input message";
```
elijahsullivan, line 13:
```suggestion
        String userInput = "foo";
```

Non-blocking: I would typically just use something like "foo" here. Most developers won't assume that there is some special behavior when "foo", "bar", "baz", etc strings are encountered in tests.

elijahsullivan: Another option would be to parameterize this test to indicate in the test that it doesn't matter what user input is entered ("foo", "bar", "baz"), we just care that it is equal to `calledWithConnection.read().open()`. You could take this a step further with "property-based testing" or "fuzz testing".

jonmanahan: I shall look into this! Thank you 😄

README.md
jonmanahan: I added both of these install steps to the prereqs of the README!

app/src/test/java/waiter/CommunicatorTest.java
```diff
@@ -0,0 +1,21 @@
...
+        ListenerMock listenerMock = new ListenerMock(userInput);
+        MessengerMock messengerMock = new MessengerMock();
+        Communicator communicator = new Communicator();
+        communicator.communicate(listenerMock, messengerMock);
```
elijahsullivan, line 17:
```suggestion
        Communicator communicator = new Communicator(listenerMock, messengerMock);
        communicator.communicate();
```

Did you consider moving the `listener` and `messenger` into the `Communicator` constructor? Do we need to be able to specify different listeners/messengers when calling `communicate`?

.idea/gradle.xml
elijahsullivan: I would do a little research on your own to see why you would/wouldn't want to include the `.idea` directory in source control. If you decide you want to exclude it, then yes I would remove the directory and files from git tracking and include it in the `.gitignore`.

app/src/test/java/waiter/CommunicatorTest.java
jonmanahan: That is a really good point, we can move them into the constructor, not sure why I didn't in the first place, thank you 😄

app/src/test/java/waiter/CommunicatorTest.java
matteeyao: Interested in this too - could you post your findings here Jon?

jonmanahan: Hmmm, so from what I understand and looked up, generally this is done by either manually creating a list of inputs or using a randomly generated list of inputs (via a library or something), picking one of those inputs randomly and then testing using that input since we are just checking for equivalence. Also I do would do this random input generation and selection in the test, right?

app/src/main/java/waiter/Awaiter.java
```diff
@@ -0,0 +1,7 @@
+package waiter;
+
+import java.io.IOException;
+
+public interface Awaiter {
+    Connection awaitClient() throws IOException;
```
mrfishball, line 6: Do we need this interface?

app/src/main/java/waiter/Transporter.java
```diff
@@ -0,0 +1,8 @@
+package waiter;
+
+
+import java.io.IOException;
+
+public interface Transporter {
+    void transport(Connection client) throws IOException;
```
mrfishball, line 7: Same as the `Awaiter` interface. Do we need this `Transporter` interface?

app/src/main/java/waiter/Awaiter.java
jonmanahan: I included this interface so I can easily mock the Listener without having to override the Listener class

.idea/gradle.xml
jonmanahan: From what I looked into, you can keep the .idea tracked if you want to, but some files within it probably shouldn't be included as they are specifically for your IntelliJ configuration (like the workspace.xml, tasks.xml) which my  .gitignore in my .idea folder takes care of. I also often saw people say they just avoided it all together, but that doesn't seem quite right to me.

app/src/main/java/waiter/Transporter.java
jonmanahan: I included this interface so I can easily mock the Messenger without having to override the Messenger class's functions

app/src/main/java/waiter/App.java
```diff
@@ -3,12 +3,16 @@
...
+import java.net.ServerSocket;
+
 public class App {
-    public String getGreeting() {
-        return "Hello World!";
-    }
 
-    public static void main(String[] args) {
-        System.out.println(new App().getGreeting());
+    public static void main(String[] args) throws IOException {
```
mrfishball, line 11: I'm not sure we want to throw here. Do we want the server to blow up upon starting up or do we want to handle any exceptions gracefully?

jonmanahan: Hmmm, Im not quite sure, but I would think we would want to handle it gracefully, would this be done via something like a try catch?

app/src/main/java/waiter/Messenger.java
```diff
@@ -0,0 +1,18 @@
+package waiter;
+
+import java.io.IOException;
+
+public class Messenger implements Transporter {
+
+    Protocol server;
+
+    Messenger(Protocol server) {
```
mrfishball, line 9: This is getting confusing to me. Somewhere in the code we're using `Awaiter server` and then we also have `ServerSocket server` and here we're using `Protocol server`. As I jump through the code I thought that all take in the same server but in reality they are completely different things. I would suggest calling them what they are, for example, `Protocol protocol`. Ultimately we want to make sure we' re using these terms in a consistent manner and avoid having one term means different things in different part of the codebase.

.idea/gradle.xml
jonmanahan: Hmmm, now the more I look into it the more I might say to not include it all together, or at least most of it. Through a couple of links I found this link that generates a git ignore file automatically depending on the IDE you put in: https://www.toptal.com/developers/gitignore which included a bit more than I thought originally, could be a useful resource 🤔

app/src/test/java/waiter/CommunicatorTest.java
```diff
@@ -0,0 +1,21 @@
...
+        Connection calledWithConnection = messengerMock.calledWith;
```
mrfishball, line 18: What is this `calledWith`?

app/src/main/java/waiter/App.java
mrfishball: Yea I think try catch should do it. At any given time we don't want our application to crash. In case where the problem can't be resolved and the user needs to re-run the application, we should let the user know that instead of just throwing an exception.

app/src/test/java/waiter/CommunicatorTest.java
jonmanahan: It is the clientConnectionMock that is set when messengerMock's transport method is called (the messengerMocks transport method is called with that clientConnectionMock), so that you an compare the user_input and the string that was set in that clientConnectionMock's write method

jonmanahan: I hope I explained this well 😄

app/src/main/java/waiter/Messenger.java
jonmanahan: Sounds good to me, I will have the names reflect what they are!

app/src/main/java/waiter/Transporter.java
mrfishball: hmm. So this interface is to make sure that the mock and the actual class have the same interface(void transport method)?

app/src/main/java/waiter/ClientConnection.java
```diff
@@ -0,0 +1,27 @@
+package waiter;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.PrintStream;
+import java.net.Socket;
+
+public class ClientConnection implements Connection {
```
mrfishball, line 9: Do we have a test for this?

jonmanahan: As a result of ClientConnection being a thin wrapper around minimal native java functionality and since I don't want to mock what I don't own, I didn't test this wrapper.

app/src/main/java/waiter/App.java
mrfishball: I think that's ok for now. Ideally we can catch a SocketExeception as well as an IOException so that we can distinguish between a connection issue or an IO issue(reading from or writing to the stream). That also means instead of throwing IOException in the `Listener`, we would throw a SocketException instead.

app/src/main/java/waiter/App.java
```diff
@@ -3,12 +3,16 @@
...
+        int port = 4424;
+        Listener listener = new Listener(new ServerSocket(port));
+        Messenger messenger = new Messenger(new EchoProtocol());
+        Communicator communicator = new Communicator();
+        communicator.communicate(listener, messenger);
```
mrfishball, line 16: What happens after the echo message is sent? does the server terminates the connection and quit?

jonmanahan: The server stops running and the program exits

app/src/main/java/waiter/App.java
jonmanahan: That does make sense, I will put in a basic try catch with a "There was a problem starting the echo server, please check your command and rerun" for now, but Ill create another PR in the future to address this too! Thank you for the explanation as well 😄

app/src/main/java/waiter/ClientConnection.java
mrfishball: There's actually quite a bit of important logic in here and there's no way for us to know if the server still works when we make changes to this class unless we manually test it which isn't ideal.

app/src/test/java/waiter/CommunicatorTest.java
```diff
@@ -0,0 +1,21 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class CommunicatorTest {
+
+    @Test
+    void establishesCommunication() throws IOException {
+        String userInput = "Doesn't matter whats here, simulating any user input message";
+        ListenerMock listenerMock = new ListenerMock(userInput);
+        MessengerMock messengerMock = new MessengerMock();
```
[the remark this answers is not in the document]
mrfishball: Suggestion: Moving the mock classes to their corresponding class to so that it's easier to update them later on.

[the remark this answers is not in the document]
jonmanahan: Like have the mock classes and the corresponding class in the same file?

app/src/main/java/waiter/ClientConnection.java
```diff
@@ -0,0 +1,27 @@
...
+
+    private final Socket client;
+
+    ClientConnection(Socket client) {
```
mrfishball, line 13: Can we call this socket instead of client?

jonmanahan: I renamed it to `clientSocket` in my name change commit I have, but I can change it to socket instead if that makes more sense

app/src/main/java/waiter/Communicator.java
```diff
@@ -0,0 +1,11 @@
+package waiter;
+
+import java.io.IOException;
+
+public class Communicator {
+    void communicate(Awaiter server, Transporter messenger) throws IOException {
+        Connection client = server.awaitClient();
```
mrfishball, line 7: Same with my other comment, let's call this `connection` instead of of `client` since somewhere in the code we're using `Socket client`.

jonmanahan: I was initially avoiding naming the variables after their types because it seemed redundant, I discovered the `var` keyword and think an ideal solution would be to use this to name them after what they are, what are your thoughts?

app/src/test/java/waiter/CommunicatorTest.java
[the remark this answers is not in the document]
mrfishball: Not the tests. I meant moving the mocks to be adjacent to their actual class for example,
```
// inside main/java/waiter/ClientConnection/ directory

ClientConnection.java
mock/

// inside the mock/ directory
ClientConnectionMock.java
```

app/src/main/java/waiter/App.java
mrfishball: Sounds good!

app/src/main/java/waiter/App.java
mrfishball: Make sense, I was thinking if we need to do some clean up like closing the connection but it looks like we're ok right now since the program terminates itself.

app/src/test/java/waiter/CommunicatorTest.java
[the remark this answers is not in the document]
jonmanahan: Hmmm that be cool, I guess my only question is why I would include the mocks in the src directory when they only really exist in the test code/directory, I appreciate all the help 😄

app/src/main/java/waiter/ClientConnection.java
mrfishball: I see. I find the name `clientSocket` confusing because my understanding is that socket exists on the server rather than on the client. Maybe it's more appropriate call a connection clientConnection. I hope that makes sense.

app/src/main/java/waiter/Communicator.java
mrfishball: Can you clarify what you meant by using the var keyword?

jonmanahan: Like using `var connection` instead of `Connection connection`, the var connection still is of type Connection in this case, just less redundant reading wise

jonmanahan: Either way is really ok to me, just thought it might look a little more clean, but maybe it then is less understandable from a reading perspective, what are your thoughts?

app/src/test/java/waiter/CommunicatorTest.java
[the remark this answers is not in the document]
mrfishball: Good question. I'll use `ClientConnection` and `ClientConnectionMock` as an example but it should apply to all classes and their mocks. This colocating is symbolic in a way that a mock is highly coupled to the actual class. For example if you update `ClientConnection.someMethodCall`, you would also need to update the mock to reflect that change. Similarly if you add a new method in ClientConnection, you'll likely need to update the mock as well. You can create a directory for all the mocks in the test directory as well but I think  colocating the mock and actual class makes updating a little easier.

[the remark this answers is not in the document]
jonmanahan: That does indeed make alot of sense, the fact that they are so heavily coupled is a good reason to keep them together, I might make that a separate PR as its more of an organizational thing, but I will definitely update later! If that sounds all good to you 😄 , I can also do it in this PR I just thought it would be better in its own.

app/src/main/java/waiter/Communicator.java
mrfishball: I see. I think I'm leaning toward Connection connection. It's quite common that the variable name is the same as the type in Java so you should be ok 😄

app/src/test/java/waiter/CommunicatorTest.java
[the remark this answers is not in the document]
mrfishball: 👍 sounds good either way. Let's make sure we capture this somewhere if you're planning on doing this in a future PR.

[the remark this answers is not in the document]
jonmanahan: Would a refactor/future PR ticket in trello be a good idea? I have it in my notes but I figured we'd want it captured somewhere we can all see it, plus I have a couple of future PR's I want to do so I can put those in tickets as well if this sounds good

app/src/main/java/waiter/ClientConnection.java
jonmanahan: This is a good point and a good way to go about testing this, I will look into injecting a reader and writer class and a reader and writer mock in tests

app/src/test/java/waiter/CommunicatorTest.java
elijahsullivan: I see a couple of options here:

1. Change the string name to something like "foo" to indicate that this is some arbitrary string.
2. Use a library like [Faker](https://github.com/DiUS/java-faker) to generate a random fake string, just be careful that you aren't introducing an intermittently failing test (sometimes referred to as a flaky test).
3. Write a parameterized test to prove that this works for a relatively small, distinct set of strings (not what I would reach for in this particular case).
4. Write a property-based test to prove that this works for a set of inputs that abide by some preconditions (i.e. all `Strings` limited to ASCII characters, with a length of 0-20, that don't contain "\n").

app/src/main/java/waiter/Communicator.java
elijahsullivan: +1 to @mrfishball I would follow the convention of `Connection connection`. I don't have a strong preference for/against using `var`/leaning on Local-Variable Type Inference unless it makes the code less readable.

.idea/gradle.xml
jonmanahan: Sounds good to me, that is what I was leaning towards as well. Thank you for the explanation too, that makes a lot of sense! I am thinking of making this change its own PR (and put it as a ticket to keep track of it) since it is more of an organizational change (maybe putting it with the mock organization change) and i feel like it doesn't really belong in this PR. What are your thoughts?

app/src/test/java/waiter/CommunicatorTest.java
jonmanahan: I have changed it to "foo" for now, I want to try and implement a property-based test as an improvement.

app/src/test/java/waiter/CommunicatorTest.java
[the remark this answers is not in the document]
elijahsullivan: @jonmanahan Thinking about this a bit more – even though this change pertains more to code organization it only really includes files that were introduced in this PR. Making the change now won't make the PR more difficult to comprehend for reviewers. I'm leaning more towards having you do this now.

I've made comments in the past about separating changes like – file naming, location changes, updated directory hierarchies – into a separate PR from feature development, but this feels different to me since the files you are organizing here were introduced as part of this feature.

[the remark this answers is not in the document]
jonmanahan: "Making the change now won't make the PR more difficult to comprehend for reviewers." I like this point and it makes alot of sense, I can see how the fact that the files were introduced in this feature makes it make sense to include this change in the current PR, as opposed to separating just because its more organization. I am also leaning towards making the change in this PR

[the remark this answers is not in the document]
jonmanahan: Check for resolved in [this commit](https://github.com/jonmanahan/waiter/pull/2/commits/b5c07b226a5e771c655351044d9ef186473cd6eb), if it resolves the issues please hit resolve conversation if possible, thank you all a ton!

README.md
jonmanahan: Check for resolved in [this commit](https://github.com/jonmanahan/waiter/pull/2/commits/e53fdbc1eec22480eae57e79838d0f1186c16f90), if it resolves the issues please hit resolve conversation if possible, thank you all a ton!

app/src/main/java/waiter/Listener.java
jonmanahan: Resolved in [this commit](https://github.com/jonmanahan/waiter/pull/2/commits/e53fdbc1eec22480eae57e79838d0f1186c16f90), if it resolves the issues please hit resolve conversation if possible, thank you all a ton!

app/src/main/java/waiter/ClientConnection.java
jonmanahan: Resolved in [this commit](https://github.com/jonmanahan/waiter/pull/2/commits/e53fdbc1eec22480eae57e79838d0f1186c16f90), if it resolves the issues please hit resolve conversation if possible, thank you all a ton!

app/src/test/java/waiter/CommunicatorTest.java
jonmanahan: Resolved in [this commit](https://github.com/jonmanahan/waiter/pull/2/commits/e53fdbc1eec22480eae57e79838d0f1186c16f90) via changing the `userinput` to "foo", but also want to continue this issue into a future PR as well to implement property-based testing. Potential ticket name -> property-based-testing ticket, thank you all for the help!

app/src/test/java/waiter/ListenerMock.java
```diff
@@ -0,0 +1,13 @@
+package waiter;
+
+public class ListenerMock implements Awaiter {
+
+    String userInput;
+
+    ListenerMock(String userInput) {
+        this.userInput = userInput;
+    }
+    public Connection awaitClient() {
```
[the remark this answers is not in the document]
jonmanahan: Resolved in [this commit](https://github.com/jonmanahan/waiter/pull/2/commits/e53fdbc1eec22480eae57e79838d0f1186c16f90), if it resolves the issues please hit resolve conversation if possible, thank you all a ton!

app/src/test/java/waiter/CommunicatorTest.java
jonmanahan: Resolved in [this commit](https://github.com/jonmanahan/waiter/pull/2/commits/e53fdbc1eec22480eae57e79838d0f1186c16f90), if it resolves the issues please hit resolve conversation if possible, thank you all a ton!

app/src/main/java/waiter/App.java
jonmanahan: Resolved in [this commit](https://github.com/jonmanahan/waiter/pull/2/commits/bf40a8e50470f025810a180272a6c38774729300), but also want to add throwing SocketExceptions in another PR as discussed above. Potential ticket name (for tracking purposes) -> add-socket-exception-catch Ticket, thank you for all the help, you guys are awesome!

.idea/gradle.xml
jonmanahan: Going to defer this to be included in a separate PR as discussed above, Potential ticket name (for tracking purposes) -> untrack-idea-directory ticket, thank you all for the help!

app/src/main/java/waiter/Messenger.java
jonmanahan: Resolved in [this commit](https://github.com/jonmanahan/waiter/pull/2/commits/3325b9eee76dbc890d1836d2e1d440ec28adb18b), if it resolves the issues please hit resolve conversation if possible, thank you all a ton!

app/src/main/java/waiter/ClientConnection.java
jonmanahan: Going to defer this to be included in a separate PR (the first/next PR), Potential ticket name (for tracking purposes) -> add-client-connection-test-coverage ticket, thank you all for the help!

app/src/main/java/waiter/App.java
jonmanahan: Just to keep track of this as well, this will be addressed in the Multiple Message ticket

app/src/main/java/waiter/Communicator.java
jonmanahan: Resolved in [this commit](https://github.com/jonmanahan/waiter/pull/2/commits/c5b65f8f2b576acc81c13d122850ed83f4052bda), if it resolves the issues please hit resolve conversation if possible, thank you all a ton!

app/src/main/java/waiter/ClientConnection.java
```diff
@@ -8,20 +8,20 @@
 
 public class ClientConnection implements Connection {
 
-    private final Socket client;
+    private final Socket clientSocket;
 
-    ClientConnection(Socket client) {
-        this.client = client;
+    ClientConnection(Socket clientSocket) {
```
mrfishball: I think this is another mismatch, In the `Listener` class we have `serverSocket` but here we're calling this `clientSocket` I think they mean the same thing in terms of socket. I would suggest either just go with the name server socket or just socket since I don't think clientSocket is a thing because sockets in general is a server term. I also noticed that we're using `import java.net.Socket` here but in `ClientConnection` we're using `import java.net.ServerSocket`. What are the difference between the two and can we just one instead of both?

jonmanahan: I fixed the `clientSocket` to just be `socket` in [this commit](https://github.com/jonmanahan/waiter/pull/2/commits/c5b65f8f2b576acc81c13d122850ed83f4052bda). In the case of the ServerSocket, we use that in listener instead of a regular socket as it has the additional feature of the `accept` method, which we need to get the socket in the `ClientConnection` above, if that makes sense.

README.md
elijahsullivan: I was able to run the project and tests with the following:

`openjdk-18.0.1.1` (installed via `asdf`)
`gradle-7.4.2` (installed via `brew`)

jonmanahan: Should I put the gradle version in the README like I did the Java?

elijahsullivan: I'll leave that up to you. The Java version is more important. You could specify that the application is known to work with X version of Gradle if you want.

app/src/main/java/waiter/ClientConnection.java
mrfishball: Make sense. Thanks for clarifying!

mrfishball (approved):
Thanks Jon!

elijahsullivan (approved):

jonmanahan:
Thank you all for all the help/guidance/answers and for reviewing this PR!

jonmanahan:
> Thanks Jon!

Thank you Steven, for all of the help/answers/good points, you helped me a lot throughout, I really appreciate it!

PR #4, opened 2022-05-10
Add Socket Exception catch

jonmanahan:
Added test coverage for catching the IOException and SocketException
Converted public classes to records wherever applicable because of IDE warnings


app/src/main/java/waiter/EchoServer.java
```diff
@@ -0,0 +1,20 @@
+package waiter;
+
+import waiter.Communicator.Reporter;
+
+import java.io.IOException;
+import java.net.SocketException;
+
+public record EchoServer(Reporter reporter) {
```
mrfishball, line 8: I'm curious what is the purpose of this `reporter`?

app/src/main/java/waiter/Communicator/Reporter.java
```diff
@@ -0,0 +1,8 @@
+package waiter.Communicator;
+
+import java.io.IOException;
+
+public interface Reporter {
```
mrfishball, line 5: I noticed there are quite a number of interfaces now and it's getting a bit difficult keeping track of what they are vs an actual class. I wonder if there are some organizationally stuff we can do to make this more manageable.

app/src/main/java/waiter/App.java
```diff
@@ -3,27 +3,19 @@
...
-            Communicator communicator = new Communicator(listener, messenger);
-            communicator.communicate();
-        }
-        catch (IOException exception) {
-            System.out.println("There was a problem starting the echo server, please check your command and rerun");
-        }
+        int port = 4424;
+        Listener listener = new Listener();
+        Messenger messenger = new Messenger(new EchoProtocol());
+        EchoServer echoServer = new EchoServer(new Communicator(listener, messenger, port));
```
mrfishball, line 18: Very nice!

app/src/main/java/waiter/Communicator/Reporter.java
jonmanahan: Hmmm, I have put each interface in a package with their associated mock (in a mock package within said package) and classes, so for example, the Reporter interface lives in a Communicator package with the Communicator class and a mock package that contains the mockCommunicator, both of which implement this interface. No class or mock implements an interface that they are not packaged with. Do you think having an interface package (like the mock) would be a good idea?

app/src/test/java/waiter/EchoServerTest.java
```diff
@@ -0,0 +1,36 @@
...
+import java.io.PrintStream;
+import java.net.SocketException;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class EchoServerTest {
+
+    private final ByteArrayOutputStream output = new ByteArrayOutputStream();
+
+    @BeforeEach
```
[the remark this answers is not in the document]
jonmanahan: Good point! I didn't even think of this but its good and good practice 😄 thank you!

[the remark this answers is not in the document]
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/4/commits/0852950c4e437c2a4fffb4aec9c5b9027d327a1f), if it looks all good and if possible please hit resolve conversation after reviewing, thank you! 😄

app/src/main/java/waiter/EchoServer.java
```diff
@@ -0,0 +1,20 @@
...
+
+    public void start() {
+
+        try {
+            reporter.communicate();
+        } catch (SocketException exception) {
+            System.out.println("Sorry, connection could not be established or has been broken, please try running the server and connecting again");
+        } catch (IOException exception) {
+            System.out.println("Sorry, there was a problem with your input, please try running the server and connecting again");
```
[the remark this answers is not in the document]
jonmanahan: That is a good point that I didn't think of and makes a lot of sense to me now that you mention it, how does `"Sorry, there was a problem with reading or writing your message, please try running the server and connecting again"` sound? or do you think it would be better (because its more general) to just say ".... problem with your message...." instead of "reading or writing"?

app/src/main/java/waiter/EchoServer.java
mrfishball: hmm. Food for thought: I wonder if the use of interface is making things more difficult to follow. I know that you mentioned testing being the main reason for the interfaces but most of them are very small, with one to two methods at most. We also don't have multiple classes implementing the same interface, I can be wrong. If you change one of the mock's method signature, your test will likely fail so we don't really need interface to enforce that. Part of me thinks we might be better off getting rid of the interfaces. Thoughts? @elijahsullivan @jonmanahan @matteeyao

app/src/main/java/waiter/EchoServer.java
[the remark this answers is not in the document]
mrfishball: Maybe something like "... an error occurred when sending/receiving a message, please try again later..."

app/src/test/java/waiter/EchoServerTest.java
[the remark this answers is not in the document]
mrfishball: Looks good!

app/src/main/java/waiter/EchoServer.java
elijahsullivan: @mrfishball +1 I share your sentiment that there might be too many interfaces.

matteeyao: Sorry guys, I'm not able to connect the dots here. But if Jon gets rid of the interfaces, which were being used to mock classes to test another class that uses the class being mocked, would he instead be creating physical test classes to test those classes being fed those mocked classes?

app/src/test/java/waiter/EchoServerTest.java
```diff
@@ -0,0 +1,44 @@
...
+import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.io.PrintStream;
+import java.io.ByteArrayOutputStream;
+
+import java.io.IOException;
+import java.net.SocketException;
+
+public class EchoServerTest {
```
elijahsullivan, line 16:
```suggestion
class EchoServerTest {
```

Non-blocking: I _think_ with JUnit5 you can depend on the default package visibility for test files.

jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/4/commits/1c0ece08eec1ec5aa9d82701db2be444022898f6), thank you!

app/src/main/java/waiter/EchoServer.java
mrfishball: In general I don't think testing should be the sole reason for introducing an interface. The purpose of an interface is to:

- make sure the contract doesn't change between multiple implementers and the consumer
- polymorphism in a way that the implementation doesn't matter

In the case where we're testing, for example:
```java
public class Foo {
    ...
    public void doSomething() {
        ...
    }
}

public class FooMock {
    ...
    public void doSomething() {
        ...
    }
}


//test

@Test
void testThatBarWorks() {
   Bar systemUnderTest = new Bar(
       new FooMock,
       ...
   );

   systemUnderTest.doStuff();
   
   assertThatBarWorks
}
```

If we change `FooMock`'s method signature of `doSomething`, I think the test will fail and so my question is do we really need interface so that we can mock a class? There is nothing stopping us from writing a FooMockTwo without an interface.

To @jonmanahan point about making the code more modular. I think writing TDD makes our code more modular. A lot of times we're thinking about how do I go about testing this class and through that process we incorporate dependency injection, single responsibility and other principle to help us write modular code so using interface is like an after the fact thing. I'm curious about the point you make about mock class overriding the original class's method.

elijahsullivan: I think my concern with the interfaces centers more around their use to facilitate the existence of mocks where we might not need them. Perhaps some of these interfaces exist where another would be more appropriate.

FWIW – the example given for an `EchoProtocol` vs an `HTTProtocol` is the kind of interface I like to see.

However, I'm particularly focused on `InputStreamer` and `OutputStreamer`. Neither of which we have tests for from what I can tell, though both perform important behavior.

In the case of the `InputStream` we could test that we call a method on the provided socket (wrapper + interface + mock here) which receives a byte stream and returns a `String`. The inverse with the `OutputStream`.

The `ClientConnection` takes an `InputStreamer` (Reader) and an `OutputStreamer` (Writer), though I wonder if it could just take a socket (again wrapped) and handle reading/writing of `byte streams -> string` and `string -> byte streams` itself. I'm less convinced of the likelihood that our Reader/Writer here will need to change since that is how messages are sent over the wire and more convinced that the socket is what is hard to test here.

Ultimately I think I'm fine keeping these if they make testing the ClientConnection easier, so long as they themselves are tested.

mrfishball (approved):
Looks good! 🚀

app/src/main/java/waiter/EchoServer.java
jonmanahan: @mrfishball That is a true, I think if we change FooMock method signature, the test does indeed fail, but this brings up the question, how would bar in this case handle receiving either a Foo and a FooMock without a common interface between them? If the constructor takes in a FooMock here in the test, then doesn't the constructor in bar have to specify it takes FooMock fooMock and then as a result can't take Foo? (if that makes sense) I could be misunderstanding here so apologies if so and I very much appreciate the analysis here regardless of the outcome!

@elijahsullivan in terms of the `InputStreamer` and `OutputStreamer` I do plan on writing tests for them and unfortunately didn't do so before hand as a result of me misunderstanding the importance in testing their behavior. I do plan on also creating a socket (wrapper mock and interface) for this and the multiple messages ticket as well. As for passing the socket and having `ClientConnection` handle the reading/writing itself, the only issue I see there is that I would either have to new up a reader and writer in the `ClientConnection` constructor, which I don't think/have read is not ideal/good, or I would have to new up the reader/writer in the methods they are used in (ie:read and write in `ClientConnection`) which would cause them to be newuped every call to them, which I don't really like either. Unless I am completely off about the new uping in a constructor, then I think it is possible to pass in just the socket and take care of the rest in `ClientConnection`. Either way I agree, the `InputStreamer` and `OutputStreamer` should be tested if they are staying and I have made note of this to keep track of it and will implement tests for them in the future!

I hope this all made sense and really can't express enough how much I appreciate all of the help and thought provoking discussions not only here but on every PR!

mrfishball: +1 @elijahsullivan yea I forgot to mention I do like the Protocol interface in a sense that we know we have a need for it because of echo vs http. I find it very hard to keep track of all the classes we already have on top of the interfaces which makes it difficult to reason about the codebase.

mrfishball: Ultimately, I'm not saying we should remove all the interface but rather pop the question of whether we need them or not for the most part. If we do need them are there ways we can organize them in a way that it's obvious that they are interface, for example, class names are usually `nouns`, so maybe we can use adjectives in cases like: 

`Reporter` -> `Reportable`
`Reader` -> `Readable`
`Transporter` -> `Transportable`

jonmanahan: I completely understand and agree with the fact that the names of the classes and the interfaces are hard to tell apart just by looking at the names, I didn't really think about it from this angle before (not having the IDE to tell you what is an interface and a class) but it makes alot of sense. I like this idea of using adjectives as the interface names alot as it allows one to discern an interface and a class even at a glance. This is a great idea in my opinion, thoughts? @matteeyao @elijahsullivan

app/src/main/java/waiter/Communicator/Reporter.java
mrfishball: I'm not sure. I've seen file organization where actually classes live in an implementation directory and interfaces live in another.

mrfishball: I was also thinking about a file structure that looks like this:
```
---Reportable/ (interface name)
     -(interface goes here)
     --Implementation/
         -(classes go here)
     --Mocks/
        -(mocks go here)
```

jonmanahan: If we implement the noun -> class, adjective -> interface, would you still recommend an organizational change like this?

jonmanahan: Hmmm, I do like this but I also have a question about it. Do you think the interface should be the directories name? Just because I feel like the implementations name is more recognizable if that makes sense. Actually the more that I think about it the more this makes sense over having the implementation name be the directory as there is only one interface to this directory but there could be multiple classes which you wouldn't want just floating around in the initial directory. I really like this organizational structure here 😄

mrfishball: Yea I'm not super sure about this structure either and I think you have a point about class name as directory name being more clear. I thought about this structure when I think about EchoProtocol and HTTPProtocol that will both implement the Protocol interface. In that case it might makes sense to put them both under `Protocol` which is the name for the interface. Because if you create separate directory for these classes, that implies that the interface will need to live alone somewhere. The thing I'm not super sure is mixing that with classes that don't implement any interface, then you end up having directories with interface names AND class names which can also be confusing.

app/src/main/java/waiter/EchoServer.java
elijahsullivan: +1 to the adjective-based interface naming.

elijahsullivan: @jonmanahan re: "I would have to new up the reader/writer in the methods they are used in (ie:read and write in ClientConnection) which would cause them to be newuped every call to them"

I agree, that doesn't seem ideal. Aren't you newing up an`InputStreamReader` and `BufferedReader` on each call to `InputStreamer.readLine()` currently?

jonmanahan: That is a really good point, Im thinking I should pull out all the new ups into listener then so they are only called once!

app/src/main/java/waiter/Communicator/Reporter.java
jonmanahan: Hmmm, would it be ok to keep it like this for now and then create a tracking ticket and address this if I finish the other tickets this sprint or next sprint? I think there is definitely a better way to organize this (some form of what we have both suggested/discussed) and want to implement it but don't want it to prevent me from continuing on things this sprint since I feel its not incredibly blocking. What are your thoughts? 😄

elijahsullivan (approved):

app/src/main/java/waiter/Communicator/Reporter.java
mrfishball: Yea definitely not blocking! I think updating the interface name to be adjectives is a good first step.

app/src/main/java/waiter/EchoServer.java
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/4/commits/dafffa3b0f3f66e269a98f51dc4d752fc438d5de), thank you!

jonmanahan: Addressed new Interface naming convention in [this commit](https://github.com/jonmanahan/waiter/pull/4/commits/513240551e9d07c20537e83fd00dbc5039e8cf20), thank you!

app/src/main/java/waiter/Communicator/Reporter.java
jonmanahan: A ticket was created to track the apps interface/class reorganization

PR #5, opened 2022-05-10
Add client connection test coverage

jonmanahan:


app/src/test/java/waiter/ClientConnectionTest.java
```diff
@@ -0,0 +1,35 @@
...
+import java.io.IOException;
+
+public class ClientConnectionTest {
+
+    @Test
+    void readsMessage() throws IOException {
+        String input = "foo";
+        InputStreamerMock inputStreamerMock = new InputStreamerMock(input);
+        OutputStreamerMock outputStreamerMock = new OutputStreamerMock();
+        ClientConnection clientConnection = new ClientConnection(inputStreamerMock, outputStreamerMock);
```
mrfishball, line 20: What do you think about inlining this:
```java
String input = "foo";

ClientConnection clientConnection = new ClientConnection(
  new InputStreamerMock(input),
  new OutputStreamerMock()
);
```

app/src/test/java/waiter/ClientConnectionTest.java
```diff
@@ -0,0 +1,35 @@
...
+        assertEquals(readMessage.open(), input);
+    }
+
+    @Test
+    void writesMessage() throws IOException {
+        String input = "foo";
+        Message outputMessage = new Message(input);
+        InputStreamerMock inputStreamerMock = new InputStreamerMock(input);
+        OutputStreamerMock outputStreamerMock = new OutputStreamerMock();
+        ClientConnection clientConnection = new ClientConnection(inputStreamerMock, outputStreamerMock);
```
mrfishball, line 31: Similarly, consider inlining here.

app/src/test/java/waiter/ClientConnectionTest.java
jonmanahan: Hmmm, I think this is a good idea, it doesn't really hurt readability and it makes it more concise, should I reflect this in `Listener` as well?

app/src/test/java/waiter/ClientConnectionTest.java
jonmanahan: That makes sense, ignore the second part of my comment above 😂

mrfishball (approved):
Left a non blocking comment. Looks good!

app/src/test/java/waiter/ClientConnectionTest.java
mrfishball: Yea sounds good to me!

app/src/test/java/waiter/ClientConnectionTest.java
mrfishball: how about this:
```java
String input = "foo";
OutputStreamerMock outputStreamerMock = new OutputStreamerMock();

ClientConnection clientConnection = new ClientConnection(
  InputStreamerMock(foo),
  outputStreamerMock
);

assertEquals(outputStreamerMock.writtenOutput, input);
```

I feel like you don't really line 28, I could be wrong though.

app/src/main/java/waiter/OutputStreamer/OutputStreamer.java
```diff
@@ -0,0 +1,20 @@
...
+public class OutputStreamer implements Writer {
+
+    private final Socket socket;
+
+    public OutputStreamer(Socket socket) {
+        this.socket = socket;
+    }
+
+    public void writeLine(String toClient) throws IOException {
+        PrintStream printStream = new PrintStream( this.socket.getOutputStream() );
```
mrfishball, line 17: Is the beginning and trailing spacing here intentional? `... new PrintStream( this.socket.getOutputStream() );`

PR #6, opened 2022-05-13
Multiple messages

jonmanahan:


app/src/main/java/waiter/Interactor/Interactive.java
```diff
@@ -0,0 +1,14 @@
+package waiter.Interactor;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+
+public interface Interactive {
```
jonmanahan, line 7: This is the wrapper for the `Socket` Java Class

Interactive was the name chosen as this is what the user uses to interact with the Server

app/src/main/java/waiter/Reactor/Reactor.java
```diff
@@ -0,0 +1,18 @@
+package waiter.Reactor;
+
+import waiter.Interactor.Interactive;
+import waiter.Interactor.Interactor;
+
+import java.io.IOException;
+import java.net.ServerSocket;
+
+public record Reactor(ServerSocket serverSocket) implements Reactive {
```
jonmanahan, line 9: This is the interface of the wrapper (Reactor) for the Socket Java Class

Reactive was the name chosen as this is what reacts to the users input, it is the response point to communicate with the user and give a response (whatever that response may be, in this case echoing back the user input)

app/src/test/java/waiter/EchoServerTest.java
```diff
@@ -30,15 +31,28 @@ void tearDown() {
...
+        assertFalse(communicatorMock.communicateWasCalled);
+    }
+
+    // This is to cover lack of integration or E2E tests, could be replaced when these test type coverages are introduced
+    @Test
+    void callsCommunicate() {
+        CommunicatorMock communicatorMock = new CommunicatorMock(false, new IOException());
+        EchoServer echoServer = new EchoServer(communicatorMock);
+        echoServer.start(4425);
+        assertTrue(communicatorMock.communicateWasCalled);
```
mrfishball, line 56: For context, Jon and I spoke about the lack of testing of this critical piece of logic. Despite the fact that it's testing a lot of the implementation, I think it's valuable to keep this test around. We might also want to look into adding e2e/integration tests.

mrfishball (approved):

PR #7, opened 2022-05-18
Client reconnection

jonmanahan:


app/src/test/java/waiter/EchoServerTest.java
```diff
@@ -52,7 +52,7 @@ void catchesSocketException() {
     void callsCommunicate() {
         CommunicatorMock communicatorMock = new CommunicatorMock(false, new IOException());
         EchoServer echoServer = new EchoServer(communicatorMock);
-        echoServer.start(4425);
+        echoServer.start(4424);
```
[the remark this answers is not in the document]
jonmanahan: It changed before because there was a issue with using the same port in sequential tests, but it was the result of finding that I wasn't cleaning up the serverSocket properly. So I changed it back to 4424 when I put the ServerSocket new up in a try with resources. I hope this made sense! 😄

app/src/main/java/waiter/EchoServer.java
```diff
@@ -11,8 +11,8 @@ record EchoServer(Reportable reportable) {
 
     public void start(int port) {
 
-        try {
-            reportable.communicate(new Reactor(new ServerSocket(port)));
+        try (ServerSocket serverSocket = new ServerSocket(port)) {
```
mrfishball, line 14: What is the difference between doing it this way versus having it instantiated in the try block?

app/src/test/java/waiter/ClientConnectionTest.java
```diff
@@ -14,15 +16,23 @@ class ClientConnectionTest {
     @Test
     void readsMessage() throws IOException {
         String input = "foo";
-        String message = new ClientConnection(new InputStreamerMock(input), new OutputStreamerMock()).read().open();
+        String message = new ClientConnection(new InteractorMock(), new InputStreamerMock(input), new OutputStreamerMock()).read();
```
mrfishball, line 19: Let's break these up into multiple lines?
```java
String message = new ClientConnection(
  new InteractorMock(),
  new InputStreamerMock(input),
  new OutputStreamerMock()
).read();
```

app/src/test/java/waiter/ClientConnectionTest.java
```diff
@@ -14,15 +16,23 @@ class ClientConnectionTest {
...
         assertEquals(message, input);
     }
 
     @Test
     void writesMessage() throws IOException {
         OutputStreamerMock outputStreamerMock = new OutputStreamerMock();
-        ClientConnection clientConnection = new ClientConnection(new InputStreamerMock("foo"), outputStreamerMock);
-        clientConnection.write(new Message("foo"));
+        ClientConnection clientConnection = new ClientConnection(new InteractorMock(), new InputStreamerMock("foo"), outputStreamerMock);
```
mrfishball, line 26: Same here

app/src/main/java/waiter/EchoServer.java
jonmanahan: When instantiated right next to the try keyword in parenthesis, it signifies it is a try with resources, so so long as the object instantiated in the try statement implements AutoClosable (has a close method, which thankfully the ServerSocket does), the objects clean up is taken care of after execution is finished via the try with resources automatically calling the objects close method.

app/src/test/java/waiter/CommunicatorTest.java
```diff
@@ -1,28 +1,33 @@
...
 
 import java.io.IOException;
 
 class CommunicatorTest {
 
+    String clientHasDisconnected;
+
+    @BeforeEach
+    void setUp() {
+        clientHasDisconnected = null;
```
mrfishball, line 20: Can we add some context as to why this needs to be null? This will help reviewer better understand our intent.

jonmanahan: Absolutely! Would doing this in the form of a comment be the way to go? Or a comment on the PR?

app/src/test/java/waiter/CommunicatorTest.java
```diff
@@ -1,28 +1,33 @@
...
+    }
+
     @Test
     void communicatesProvidedInput() throws IOException {
-        ListenerMock listenerMock = new ListenerMock("foo");
+        String[] clientRequests = {"curl foo1", "curl foo2", "curl foo3", "curl foo4"};
+        ListenerMock listenerMock = new ListenerMock(new String[]{"foo", clientHasDisconnected});
```
mrfishball, line 26: What is the significance of this?

jonmanahan: Hmmm, are you asking this about the strings? Im not sure which part, Im sorry ☹️

app/src/test/java/waiter/MessengerTest.java
```diff
@@ -5,17 +5,40 @@
...
+
+    String clientHasDisconnected;
+
+    @BeforeEach
+    void setUp() {
+        clientHasDisconnected = null;
+    }
+
+    @Test
+    void echosSingleMessage() throws IOException {
```
mrfishball, line 24: Can we also apply similar formatting to this test? Personally I like to group test setup into `input`, `exercise the function under test` and `assert`, each session with new lines separating them. This makes the test easier to reason about.

app/src/test/java/waiter/CommunicatorTest.java
```diff
@@ -1,28 +1,33 @@
...
         MessengerMock messengerMock = new MessengerMock();
         Communicator communicator = new Communicator(listenerMock, messengerMock);
-        ReactorMock reactor = new ReactorMock(5);
+        ReactorMock reactor = new ReactorMock(clientRequests);
         communicator.communicate(reactor);
-        Connectable clientConnectionMock = messengerMock.calledWith;
-        assertEquals("foo", clientConnectionMock.read().open());
-        assertEquals(reactor.currentNumberOfEchos, 5);
+        assertEquals(clientRequests.length, reactor.numberOfAcceptedClients);
```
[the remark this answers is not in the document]
jonmanahan: Its not so much client input as it is client requests to connect to the server, so it is trying to say that 4 separate connections (via 4 curl -v telnet://localhost:4424) were sent one after another (one connects, echoes a message and then disconnects, then the next connection is made and so on). It is done for times and then the server is set to close, in which case it exits the for loop, I hope this makes sense!

app/src/test/java/waiter/MessengerTest.java
jonmanahan: Absolutely, I completely forgot this general test format, something I should definitely remember because it is good to do always!

app/src/test/java/waiter/MessengerTest.java
```diff
@@ -5,17 +5,40 @@
...
+        String[] clientInputs = {"foo", clientHasDisconnected};
+        String[] expected = Arrays.copyOfRange(clientInputs, 0, clientInputs.length - 1);
+        ClientConnectionMock clientConnectionMock = new ClientConnectionMock(clientInputs);
+        EchoProtocolMock echoProtocolMock = new EchoProtocolMock();
+        Messenger messenger = new Messenger(echoProtocolMock);
+        messenger.transport(clientConnectionMock);
+        assertArrayEquals(expected, clientConnectionMock.echoedInputs);
```
mrfishball, line 31: Maybe it's easier if we just do `assertArrayEquals({"foo"}, clientConnectionMock.echoedInputs);` and we don't need to use `Arrays.copyOfRange` up there.

app/src/test/java/waiter/ClientConnectionTest.java
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/7/commits/713d471c9fe377388299139e90627d4683daca6c), formatted to have the input -> exercise -> assert clearly separated 👍

app/src/test/java/waiter/ClientConnectionTest.java
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/7/commits/713d471c9fe377388299139e90627d4683daca6c), formatted to have the input -> exercise -> assert clearly separated 👍

app/src/test/java/waiter/ClientConnectionTest.java
```diff
@@ -14,15 +16,23 @@ class ClientConnectionTest {
...
+        clientConnection.write("foo");
         assertEquals(outputStreamerMock.writtenOutput, "foo");
     }
+
+    @Test
+    void closesSocket() throws IOException {
+        InteractorMock interactorMock = new InteractorMock();
+        ClientConnection clientConnection = new ClientConnection(interactorMock, new InputStreamerMock("foo"), new OutputStreamerMock());
```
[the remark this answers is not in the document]
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/7/commits/713d471c9fe377388299139e90627d4683daca6c), formatted to have the input -> exercise -> assert clearly separated 👍

app/src/test/java/waiter/CommunicatorTest.java
```diff
@@ -1,28 +1,33 @@
...
+
+    @BeforeEach
+    void setUp() {
+        clientHasDisconnected = null;
+    }
+
     @Test
     void communicatesProvidedInput() throws IOException {
-        ListenerMock listenerMock = new ListenerMock("foo");
+        String[] clientRequests = {"curl foo1", "curl foo2", "curl foo3", "curl foo4"};
```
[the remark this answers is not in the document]
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/7/commits/713d471c9fe377388299139e90627d4683daca6c), formatted to have the input -> exercise -> assert clearly separated 👍

app/src/main/java/waiter/EchoServer.java
mrfishball: ah I see, thanks for clarifying!

app/src/test/java/waiter/MessengerTest.java
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/7/commits/713d471c9fe377388299139e90627d4683daca6c), formatted to have the input -> exercise -> assert clearly separated 👍

app/src/test/java/waiter/MessengerTest.java
```diff
@@ -5,17 +5,40 @@
...
+    }
+
     @Test
-    void echosMessage() throws IOException {
-        ClientConnectionMock clientConnectionMock = new ClientConnectionMock("Sample user message");
+    void echosManyMessages() throws IOException {
```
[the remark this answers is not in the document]
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/7/commits/713d471c9fe377388299139e90627d4683daca6c), formatted to have the input -> exercise -> assert clearly separated 👍

app/src/test/java/waiter/CommunicatorTest.java
mrfishball: I like the idea of a comment.

app/src/test/java/waiter/MessengerTest.java
jonmanahan: We can do that, I thought the assert might be harder to read that way, but the more I think about it, although the assert is easier to read, the expected assignment looks like a bit of a mess. Ill make this change now, thank you! 😄

app/src/test/java/waiter/CommunicatorTest.java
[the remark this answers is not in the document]
mrfishball: Make sense. I think the confusion is coming from the name of this test which is `communicatesProvidedInput`. This test is more like `shouldAllowReconnect` or something like that if I'm understanding correctly?

app/src/test/java/waiter/CommunicatorTest.java
jonmanahan: Ahhhh, the (kinda) rare case of using a comment and my first comment in the project I believe, exciting! Would a comment saying something like `"The echo servers reader uses null to determine when the client should be closed to allow for reconnection/another connection"` or is that too long? Should it more be something like `"Needs to be null to get out of the client I/O loop"`

mrfishball (approved):
Looks good, feel free to merge once you've addressed the formatting. Thanks Jon!

app/src/test/java/waiter/CommunicatorTest.java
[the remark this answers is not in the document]
jonmanahan: That is a good point, I think the `communicatesProvidedInput` test name is outdated, `shouldAllowReconnect` makes more sense but I think is still a little misleading as it can allow for both reconnection and another connection if that makes sense, so maybe `shouldAllowSequentialConnections` or something, if that reasoning makes sense 😄

app/src/test/java/waiter/CommunicatorTest.java
mrfishball: No worries, I think you answered it in your other comment so it's all good!

jonmanahan:
> Looks good, feel free to merge once you've addressed the formatting. Thanks Jon!

Of course! As always thank you guys for the review! 🥳 You all are the cats pajamas! 🐱

app/src/test/java/waiter/CommunicatorTest.java
mrfishball: I was think about something like "This needs to be null because reading the input stream returns null when the sockets are disconnected which signals the client is disconnected".

app/src/test/java/waiter/CommunicatorTest.java
```diff
@@ -1,28 +1,35 @@
...
-        communicator.communicate(reactor);
-        Connectable clientConnectionMock = messengerMock.calledWith;
-        assertEquals("foo", clientConnectionMock.read().open());
-        assertEquals(reactor.currentNumberOfEchos, 5);
+        String[] clientRequests = {"curl foo1", "curl foo2", "curl foo3", "curl foo4"};
+        ReactorMock reactor = new ReactorMock(clientRequests);
+
+        new Communicator(
+                new ListenerMock(new String[]{"foo", clientHasDisconnected}),
+                new MessengerMock()
```
mrfishball, line 30: Extra whitespace?

app/src/test/java/waiter/CommunicatorTest.java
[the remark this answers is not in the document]
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/7/commits/97d783bb1cde52ca12d9dc231d05bf967e161ca5), fixed the test name, some more test format and added a comment to the null variable declaration.

app/src/test/java/waiter/MessengerTest.java
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/7/commits/97d783bb1cde52ca12d9dc231d05bf967e161ca5), fixed the test name, some more test format and added a comment to the null variable declaration.

PR #10, opened 2022-05-20
Add property based testing where appropriate, Update README

jonmanahan:


app/src/test/java/waiter/MessengerTest.java
```diff
@@ -21,23 +27,25 @@ void setUp() {
         clientHasDisconnected = null;
     }
 
-    @Test
-    void echosSingleMessage() throws IOException {
-        String[] clientInputs = {"foo", clientHasDisconnected};
+    @Property(tries = 5)
```
elijahsullivan, line 30: What does the "tries = 5" give us here?

app/src/test/java/waiter/MessengerTest.java
```diff
@@ -21,23 +27,25 @@ void setUp() {
...
 
-        assertArrayEquals(new String[]{"foo"}, clientConnectionMock.echoedInputs);
+        assertArrayEquals(new String[]{userInput}, clientConnectionMock.echoedInputs);
     }
 
-    @Test
-    void echosManyMessages() throws IOException {
-        String[] clientInputs = {"foo", "bar", "baz", clientHasDisconnected};
+    @Property(tries = 5)
+    void echosManyMessages(@ForAll @Size(3) List<@AlphaChars @StringLength(min = 1, max = 10) String> userInputs) throws IOException {
```
elijahsullivan, line 41: Is the length of the String relevant here? Do we see issues if we allow this length to be unbounded?

app/src/test/java/waiter/MessengerTest.java
elijahsullivan: It looks like the default amount is 1,000 tries if an amount isn't specified. Did you notice that the tests were slow when it was running this many iterations?

jonmanahan: I don't think they were running slow (just tried it again, all tests took 2 seconds total), just thought 1000 was overkill initially, but its also better coverage, what are your thoughts? If its not taking a long time, I don't think 1000 tries is a bad idea, then again as the number of property based tests expand it could get worse. 🤔

elijahsullivan: I think using the default is fine for now if it isn't slowing the tests down considerably.

elijahsullivan (approved):

app/src/test/java/waiter/MessengerTest.java
jonmanahan: Default used in [this commit](https://github.com/jonmanahan/waiter/pull/10/commits/c4465f83b04314b53d2b69d6256f3eb0c3844b9b)👍

PR #12, opened 2022-06-01
Simultaneous client messaging

jonmanahan:


app/src/test/java/waiter/CommunicatorTest.java
```diff
@@ -24,14 +28,31 @@ void setUp() {
...
+        ReactorMock reactor = new ReactorMock(clientRequests);
+        ThreadGeneratorMock threadGeneratorMock = new ThreadGeneratorMock(numberOfThreadsToGenerate);
+
+        new Communicator(
+                threadGeneratorMock,
+                new ListenerMock(new String[]{"foo", clientHasDisconnected}),
+                new MessengerMock()
+        ).communicate(reactor);
+
+        assertEquals(clientRequests.length * numberOfThreadsToGenerate, reactor.numberOfAcceptedClients * threadGeneratorMock.numberOfThreadsGenerated);
```
matteeyao, line 56: Having trouble understanding why these should equal as I don't know what each represents. Could you perhaps place a variable name on them e.g.

someDescriptiveVariableName = clientRequests.length * numberOfThreadsToGenerate
anotherDescriptiveVariableName = reactor.numberOfAcceptedClients * threadGeneratorMock.numberOfThreadsGenerated
assertEquals(someDescriptiveVariableName, anotherDescriptiveVariableName)

jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/12/commits/beebfdc3e2c9d8f1314123c5ec4d473e71a203b0), Thank you :D

matteeyao (approved):
LGTM. Thanks for the changes! If this is blocking, maybe merge? But prob be better to wait for some more feedback from Eli or Steven.

Great work as always!

app/src/main/java/waiter/Threadable/mock/ThreadGeneratorMock.java
```diff
@@ -0,0 +1,21 @@
...
+    private final int numberOfThreadsToGenerate;
+    public int numberOfThreadsGenerated;
+
+    public ThreadGeneratorMock(int numberOfThreadsToGenerate) {
+        this.numberOfThreadsToGenerate = numberOfThreadsToGenerate;
+        this.numberOfThreadsGenerated = 0;
+    }
+
+    public void generate(Runnable runnable) {
+        while(this.numberOfThreadsGenerated < this.numberOfThreadsToGenerate) {
```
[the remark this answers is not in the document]
jonmanahan: Hmmm, Id say we don't need it per say, it just allows for a little more verbose test coverage as now the mock is configurable with the number of times it can be run and therefore can use property based testing. I can make it just run a hard coded number more than one (2 or 3) number of times instead, but I thought this was a little more verbose when it came to triangulation for the CommunicatorTest. I hope I explained this well, What are your thoughts? 😄

app/src/test/java/waiter/CommunicatorTest.java
```diff
@@ -24,14 +28,33 @@ void setUp() {
...
+                new ThreadGeneratorMock(numberOfThreadsToGenerate),
                 new ListenerMock(new String[]{"foo", clientHasDisconnected}),
                 new MessengerMock()
         ).communicate(reactor);
 
         assertEquals(clientRequests.length, reactor.numberOfAcceptedClients);
     }
+
+    @Property
+    void shouldAllowMultipleConnections(@ForAll @IntRange(min = 1, max = 10) Integer numberOfThreadsToGenerate) throws IOException {
```
mrfishball, line 45: How is this test any different from the one above? How do we discern between sequential connections and multiple connections? For example, looking at the setup, the way client requests are made is the same but in reality, it's very different. Maybe this test should be `create as many threads as there are connections`?

app/src/main/java/waiter/Threadable/mock/ThreadGeneratorMock.java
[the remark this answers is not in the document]
mrfishball: The generate method is called in a while loop in `Communicator` so we can just have the injected `ThreadGeneratorMock` increment a variable `numberofThread` and increment that every time `generate` is called?

app/src/test/java/waiter/CommunicatorTest.java
jonmanahan: Hmmm, I think I see what you mean, the name is not necessarily accurate/properly descriptive 🤔, @mrfishball are you available for a quick pair to discuss this?

mrfishball (approved):
Left some non blocking comments otherwise looks good!

app/src/test/java/waiter/CommunicatorTest.java
mrfishball: Not only that, I think what we really want to test is that it spins up 2 threads if there are 2 connections/requests.

PR #14, opened 2022-06-09
Simple get

jonmanahan:


app/src/main/java/waiter/Router.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class Router {
+
+    Map<String, Map<String, String>> routes;
+
+    public Router() {
+        this.routes = Map.ofEntries(
+                Map.entry("GET /simple_get", Map.ofEntries(
+                        Map.entry("status", "200 OK"),
+                        Map.entry("headers", "Accept-Language: en-US\r\nContent-Length: 0"),
```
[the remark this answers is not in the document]
jonmanahan: I put it in to show format of the headers entry, but there is no necessity for it, I can take it out as a result!

app/src/main/java/waiter/Router.java
```diff
@@ -0,0 +1,30 @@
...
+                        Map.entry("body", "")
+                )),
+                Map.entry("GET /simple_get_with_body", Map.ofEntries(
+                        Map.entry("status", "200 OK"),
+                        Map.entry("headers", "Accept-Language: en-US\r\nContent-Length: 11"),
+                        Map.entry("body", "Hello world")
```
elijahsullivan, line 19: I'm starting to see some patterns emerge here for setting data on the response.

Also, the `Content-Length` value could be calculated.

jonmanahan: Should I handle this pattern differently? Will try to calculate the content length 😄 thank you!

app/src/test/java/waiter/HttProtocolTest.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @Test
+    void simpleGetWithBodyResponse() {
+        String fromClient = "GET /simple_get_with_body HTTP/1.1";
+        HttProtocol httProtocol = new HttProtocol(new RequestParser(), new Router(), new ResponseBuilder());
+
+        String response = httProtocol.serve(fromClient);
+
+        assertTrue(response.contains("HTTP/1.1 200 OK"));
+        assertTrue(response.contains("\r\n\r\nHello world"));
```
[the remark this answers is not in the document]
jonmanahan: I feel like the `RouterTest` covers the headers test coverage and the `ResponseBuilderTest` covers the format of the response, but I can test them both here as a whole (I do like this idea) if that makes sense too 😄

app/src/main/java/waiter/Router.java
elijahsullivan: I think this pattern is fine for now, but I would consider how to extract some of the repetition as you continue to build this out.

jonmanahan: Will do, Ill look into it!

app/src/main/java/waiter/Router.java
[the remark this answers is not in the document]
elijahsullivan: Gotcha. I would remove it since it isn't strictly necessary yet. You'll build out more of the header formatting soon enought ot meet future functional requirements.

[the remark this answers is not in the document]
jonmanahan: Sounds good 😄

app/src/main/java/waiter/Router.java
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/14/commits/3e780f258f08e75533484a8eec8eec26e229510b), Since it made it easier to calculate the content length, I did a little extraction on the routes construction too, let me know if anything looks worse! Thank you 😄

app/src/main/java/waiter/Router.java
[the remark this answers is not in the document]
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/14/commits/3e780f258f08e75533484a8eec8eec26e229510b)

app/src/main/java/waiter/Router.java
```diff
@@ -0,0 +1,31 @@
...
+    Map<String, Map<String, String>> routes;
+
+    public Router() {
+        this.routes = Map.ofEntries(
+                Map.entry("GET /simple_get", setRoute("200 OK", "Content-Length:", "")),
+                Map.entry("GET /simple_get_with_body", setRoute("200 OK", "Content-Length:", "Hello world"))
+        );
+    }
+
+    private Map<String, String> setRoute(String status, String headers, String body) {
```
elijahsullivan, line 16: This method is named `setRoute`, but it takes the arguments `status`, `headers`, and `body` – all of which seem to be more concerned with the details of a response rather than the route with which they are associated.

elijahsullivan (approved):
LGTM. I just have one outstanding comment about a method name.

I see some areas where your code will likely need to change to handle future server behavior, but we'll let those features drive out the necessary code changes.

app/src/main/java/waiter/Router.java
jonmanahan: That is a really good point, I agree, maybe something like `setResponseFields`?

jonmanahan:
> LGTM. I just have one outstanding comment about a method name.
> 
> I see some areas where your code will likely need to change to handle future server behavior, but we'll let those features drive out the necessary code changes.

Thank you for the review 😄, I will for sure be changing that method name!

Sounds like a plan to me! Thank you again 😄

app/src/main/java/waiter/Router.java
jonmanahan: Im gonna go with this for now, if we can find a better one I can always change it  in the future 😄

jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/14/commits/79f9d4fd94a507209a16c6c92dc3b0231800b339)

PR #15, opened 2022-06-13
Refactor read and write to read and write bytes

jonmanahan:


app/src/main/java/waiter/Connectable/ClientConnection.java
```diff
@@ -20,11 +21,16 @@ public ClientConnection(Socket socket) throws IOException {
     }
 
     public String read() throws IOException {
-        return this.bufferedReader.readLine();
+        StringBuilder requestContents = new StringBuilder();
+        while(!requestContents.toString().contains("\r\n")) {
```
matteeyao, line 25: Not too sure if you've read [Law of Demeter](https://medium.com/vattenfall-tech/the-law-of-demeter-by-example-fd7adbf0c324), but I think the concept applies here (could be wrong).

So a lot of the newly changed code has `.` operations followed by more `.` operations. For someone like me who's terrible at reading/understanding code, this makes it really difficult to comprehend whats going on.

Perhaps you could set `requestContents.toString()` to a variable that makes me go "ah, so that's what that is" - like `formattedRequestAsString = requestContents.toString()` and then do `formattedRequestAsString.contains("\r\n")`.

Just makes for better readability, but Eli and Steven, please feel free to lmk if that doesn't apply here.

app/src/main/java/waiter/Connectable/ClientConnection.java
```diff
@@ -20,11 +21,16 @@ public ClientConnection(Socket socket) throws IOException {
...
+            requestContents.append((char) this.bufferedReader.read());
+        }
+
+        return requestContents.toString().strip();
     }
 
     public void write(String toClient) throws IOException {
-        this.printStream.print(toClient);
+        this.printStream.write(toClient.getBytes(StandardCharsets.UTF_8));
```
matteeyao, line 33: Possible breaking up of all the `.`s?

app/src/main/java/waiter/Connectable/ClientConnection.java
elijahsullivan, line 25: So this will stop reading input once a CRLF is encountered? I think you might bump into issues with this in the future if I'm understanding this correctly, but if it is working with the current server specs I'm fine with approving this as is.

elijahsullivan (approved):

app/src/main/java/waiter/Connectable/ClientConnection.java
jonmanahan: It reads until it hits the first CRLF, so it reads the `GET  /simple_get HTTP/1.1` and then encounters a "\r\n" so it stops reading and then strips the line to work properly currently. I'll keep it like this for now like you said since it is indeed working, but I will also keep an eye out for future problems I might encounter. Thank you for both the help and the heads up 😄

app/src/main/java/waiter/Connectable/ClientConnection.java
```diff
@@ -20,11 +21,16 @@ public ClientConnection(Socket socket) throws IOException {
     }
 
     public String read() throws IOException {
-        return this.bufferedReader.readLine();
+        StringBuilder requestContents = new StringBuilder();
+        while(!requestContents.toString().contains("\r\n")) {
+            requestContents.append((char) this.bufferedReader.read());
```
mrfishball, line 26: What does `requestContents` look like at the end? Are we trying to read the header of the request here?

app/src/test/java/waiter/ClientConnectionTest.java
```diff
@@ -20,7 +20,7 @@ class ClientConnectionTest {
 
     @Property
     void readsMessage(@ForAll @AlphaChars @NotBlank String userInput) throws IOException {
-        var byteInputStream = new ByteArrayInputStream(userInput.getBytes(StandardCharsets.UTF_8));
+        var byteInputStream = new ByteArrayInputStream((userInput + "\r\n").getBytes(StandardCharsets.UTF_8));
```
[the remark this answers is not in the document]
jonmanahan: Addressed in [this commit here](https://github.com/jonmanahan/waiter/pull/15/commits/da4cec5984fc9f2abea5b44b73e92d5d7778396c), used an intermittent variable `submittedUserInput = userInput + "\r\n"`

app/src/main/java/waiter/Connectable/ClientConnection.java
jonmanahan: Addressed in [this commit here](https://github.com/jonmanahan/waiter/pull/15/commits/da4cec5984fc9f2abea5b44b73e92d5d7778396c), used an intermittent variable `bytesToWriteToClient = toClient.getBytes(StandardCharsets.UTF_8)`

app/src/main/java/waiter/Connectable/ClientConnection.java
jonmanahan: Addressed in [this commit here](https://github.com/jonmanahan/waiter/pull/15/commits/da4cec5984fc9f2abea5b44b73e92d5d7778396c)

app/src/main/java/waiter/Connectable/ClientConnection.java
mrfishball: I see. Thanks!

PR #16, opened 2022-06-14
Circle ci setup

jonmanahan:


.circleci/config.yml
```diff
@@ -0,0 +1,46 @@
...
+    docker:
+      - image: "cimg/openjdk:17.0.3"
+    steps:
+      - checkout
+      - run:
+          command: gradle run
+          background: true
+      - run: git submodule update --init --recursive
+      - ruby/install:
+          version: '2.6.8'
```
mrfishball, line 28: During our pairing, I think we found out that the ruby version `2.5.1` specified for `http_server_spec` is out of date and certain gem requires a higher version in order to work. Do you want to update the `http_server_spec` to use version `2.6.8` instead?

.circleci/config.yml
```diff
@@ -0,0 +1,46 @@
...
+      - run:
+          command: |
+            cd http_server_spec
+            gem install bundler
+            bundle install --path vendor/bundle
+            sudo apt install lsof
```
mrfishball, line 34: Not suggesting a change. I'm curious why `lsof` is needed.

mrfishball: Ignore my last comment. You mentioned it before and I forgot. For the other mentors:
Screen Shot 2022-06-14 at 5 36 01 PM

.circleci/config.yml
```diff
@@ -0,0 +1,46 @@
...
+workflows:
+  build_and_test:
+    jobs:
+      - build
+      - unitTests:
+          requires:
+            - build
+      - acceptanceTests:
+          requires:
+            - build
```
mrfishball, line 46: Not related to the change. Can you try running the pipeline with a failing acceptance test?

mrfishball (approved):
LGTM! 🎸

.circleci/config.yml
jonmanahan: This is a really good idea, I will for sure try running it with a failing test and post the results here!

.circleci/config.yml
jonmanahan: Pairing with Eli at 11 EST to fix this and some commits I made on the repo by mistake 😶

.circleci/config.yml
jonmanahan: Acceptance tests fail stopping circleci [in this commit](https://github.com/jonmanahan/waiter/pull/16/commits/881f234bc04b17366de37d3e832bce3f1d6b3fd0), unit tests fail stopping circle ci [in this commit](https://github.com/jonmanahan/waiter/pull/16/commits/c9494704faa34c9a0522cdecfccba11334979681)

mrfishball: Looks like the build step also runs unit test and will fail if any unit tests fails. Maybe we should just have the build step be `build and unit test` then we don't need a separate step for unit tests.

jonmanahan: will do!

PR #17, opened 2022-06-15
Add HEAD /simple_get route and tests

jonmanahan:


matteeyao (approved):
Looks good ✅

app/src/main/java/waiter/Connectable/ClientConnection.java
```diff
@@ -21,13 +21,13 @@ public ClientConnection(Socket socket) throws IOException {
     }
 
     public String read() throws IOException {
-        StringBuilder requestStartLineBuilder = new StringBuilder();
-        while(notEndOfLine(requestStartLineBuilder)) {
-            requestStartLineBuilder.append((char) this.bufferedReader.read());
+        StringBuilder stringBuilder = new StringBuilder();
```
mrfishball, line 24: 👍 for using `StringBuilder` instead. Much easier to read.

app/src/main/java/waiter/Connectable/ClientConnection.java
```diff
@@ -21,13 +21,13 @@ public ClientConnection(Socket socket) throws IOException {
...
+        while(notEndOfLine(stringBuilder)) {
+            stringBuilder.append((char) this.bufferedReader.read());
         }
 
-        String requestStartLine = requestStartLineBuilder.toString();
-        return requestStartLine.strip();
+        String readInString = stringBuilder.toString();
+        return readInString.strip();
```
mrfishball, line 30: Consider inlining this to:
```java
return stringBuilder.toString().strip();
```

jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/17/commits/57edb196601a5e57e7dc93da5576472860a385da)

app/src/main/java/waiter/Router.java
```diff
@@ -1,23 +1,31 @@
...
+import java.util.AbstractMap;
 import java.util.Map;
 
 public class Router {
 
     private final Map<String, Map<String, Map<String, String>>> routes;
     private final String noMethod = "No Method";
     private final String noUrl = "No Url";
+    Map.Entry<String, Map<String, String>> simpleGet = new AbstractMap.SimpleEntry<>("/simple_get", setResponseFields("200 OK", "Content-Length:", ""));
+    Map.Entry<String, Map<String, String>> noUrlFound = new AbstractMap.SimpleEntry<>(noUrl, setResponseFields("500 Internal Server Error", "Content-Length:", ""));
```
mrfishball, line 12: What is the status code we're returning for `noUrlFound`? I don't think `500 Internal Server Error` is appropriate here. https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404

mrfishball: I'm curious what will happen if I try to access a URL that isn't registered? For example: `/blah`. Would I get a 500 internal server error?

jonmanahan: > What is the status code we're returning for `noUrlFound`? I don't think `500 Internal Server Error` is appropriate here. https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404

I am using 500 for right now as a general place holder because responses such as 404 not found are future tickets/implementations, so I didn't want to do future work now if that makes sense

jonmanahan: > I'm curious what will happen if I try to access a URL that isn't registered? For example: `/blah`. Would I get a 500 internal server error?

It would indeed return a 500 internal server error

app/src/main/java/waiter/Router.java
```diff
@@ -1,23 +1,31 @@
...
 
     public Router() {
         this.routes = Map.ofEntries(
```
mrfishball, line 15: If I want to add a new route, do I have to modify this `Router` class?

app/src/main/java/waiter/Router.java
mrfishball: I see. If this route is not needed, is there a reason we should keep it?

app/src/main/java/waiter/Router.java
jonmanahan: As of right now yes, but Im looking into how to make it more modular with the use of the Builder Pattern, I was thinking this would be a refactor PR on its own, to make this just getting it working and then having a separate refactor, but I could try to implement it in this PR too

mrfishball: 👍 I think a separate refactoring ticket makes sense. I'd also throw in inspirations from other web frameworks on how to handle adding new routes/resources for example [Sinatra](http://sinatrarb.com/intro.html).

app/src/main/java/waiter/Router.java
jonmanahan: As of right now, If I dont have it as this "default" then when the server gets a request it is not familiar with it would not be able to handle it in any form, resulting in the server hanging up

mrfishball: I see. I'd suggest updating that to return 404 instead of 500. If 404 not found is slated for the future but we realized it's needed for the server to work then we should just implement it based on the HTTP specification. Right now we're in a weird spot where it's can sort of handle not found but not quite right. @matteeyao @elijahsullivan @jonmanahan thoughts?

app/src/test/java/waiter/HttProtocolTest.java
```diff
@@ -27,4 +27,14 @@ void simpleGetWithBodyResponse() {
         assertTrue(response.contains("HTTP/1.1 200 OK"));
         assertTrue(response.contains("\r\n\r\nHello world"));
     }
+
+    @Test
+    void headSimpleGetResponse() {
+        String fromClient = "HEAD /simple_get HTTP/1.1";
+        HttProtocol httProtocol = new HttProtocol(new RequestParser(), new Router(), new ResponseBuilder());
```
mrfishball, line 34: Formatting nitpick:
```java
HttProtocol httProtocol = new HttProtocol(
  new RequestParser(),
  new Router(),
  new ResponseBuilder()
);
```

app/src/main/java/waiter/Router.java
matteeyao: Totally agree w/ Steven. If the change does end up breaking your acceptance test, I would keep 500 to account for whatever route is being tested, but change default route to 404 to capture every other unidentified route.

app/src/test/java/waiter/HttProtocolTest.java
```diff
@@ -27,4 +27,14 @@ void simpleGetWithBodyResponse() {
         assertTrue(response.contains("HTTP/1.1 200 OK"));
         assertTrue(response.contains("\r\n\r\nHello world"));
     }
+
+    @Test
+    void headSimpleGetResponse() {
```
mrfishball, line 32: Can we update the description to say `headSimpleGetReturnsOKWithoutBody` or something a along those line?

app/src/test/java/waiter/HttProtocolTest.java
```diff
@@ -27,4 +27,14 @@ void simpleGetWithBodyResponse() {
...
+        String fromClient = "HEAD /simple_get HTTP/1.1";
+        HttProtocol httProtocol = new HttProtocol(new RequestParser(), new Router(), new ResponseBuilder());
+
+        String response = httProtocol.serve(fromClient);
+
+        assertTrue(response.contains("HTTP/1.1 200 OK"));
```
mrfishball, line 38: Is there a way to assert that the response has no body?

app/src/main/java/waiter/Router.java
```diff
@@ -1,23 +1,31 @@
...
 
+import java.util.AbstractMap;
 import java.util.Map;
 
 public class Router {
 
     private final Map<String, Map<String, Map<String, String>>> routes;
     private final String noMethod = "No Method";
     private final String noUrl = "No Url";
+    Map.Entry<String, Map<String, String>> simpleGet = new AbstractMap.SimpleEntry<>("/simple_get", setResponseFields("200 OK", "Content-Length:", ""));
```
mrfishball, line 11: Nitpick to make it a bit easier to read:
```java
Map.Entry<String, Map<String, String>> simpleGet = new AbstractMap.SimpleEntry<>(
  "/simple_get",
  setResponseFields(...)
);

Map.Entry<String, Map<String, String>> noUrlFound = new AbstractMap.SimpleEntry<>(
  noUrl,
  setResponseFields(...)
);
```

app/src/main/java/waiter/Router.java
```diff
@@ -1,23 +1,31 @@
...
+    Map.Entry<String, Map<String, String>> noUrlFound = new AbstractMap.SimpleEntry<>(noUrl, setResponseFields("500 Internal Server Error", "Content-Length:", ""));
 
     public Router() {
         this.routes = Map.ofEntries(
                 Map.entry("GET", Map.ofEntries(
-                        Map.entry("/simple_get", setResponseFields("200 OK", "Content-Length:", "")),
+                        simpleGet,
                         Map.entry("/simple_get_with_body", setResponseFields("200 OK", "Content-Length:", "Hello world")),
```
mrfishball, line 18: Any reason we inline this route but not the others?

jonmanahan: The inlined ones were not used repeatedly so I didn't think pulling them out would be necessary (open to change though)

app/src/main/java/waiter/Router.java
mrfishball: Also for future consideration, create a wrapper around AbstractMap so that we don't need to expose the underlying type of `Map.Entry<String, Map<String, String>>` and it can just be:
```Java
Route simpleGet = new Route("/simple_get", setReponseFields(...));
```

app/src/test/java/waiter/HttProtocolTest.java
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/17/commits/01218f48d001e0c113e1b007fc2d2fe63f0aad2b)

app/src/test/java/waiter/HttProtocolTest.java
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/17/commits/01218f48d001e0c113e1b007fc2d2fe63f0aad2b)

app/src/test/java/waiter/HttProtocolTest.java
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/17/commits/01218f48d001e0c113e1b007fc2d2fe63f0aad2b)

app/src/test/java/waiter/RouterTest.java
```diff
@@ -7,7 +7,7 @@
 public class RouterTest {
 
     @Test
-    void simpleGetRoute() {
+    void getSimpleGetRoute() {
```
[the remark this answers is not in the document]
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/17/commits/01218f48d001e0c113e1b007fc2d2fe63f0aad2b)

.circleci/config.yml
```diff
@@ -27,6 +27,7 @@ jobs:
             bundle install --path vendor/bundle
             sudo apt install lsof
             bundle exec spinach --tags simple-get
```
mrfishball, line 29: I think you can do `bundle exec spinach --tags @simple-get, @head-to-simple-get `.

jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/17/commits/6fd747773a9adb7e5d4ef5e1a21cfb63ff761dcb)

app/src/main/java/waiter/App.java
```diff
@@ -15,7 +15,7 @@ public static void main(String[] args) {
 
         int port = 5000;
         Listener listener = new Listener();
-        Messenger messenger = new Messenger(new HttProtocol(new RequestParser(), new Router(), new ResponseBuilder()));
+        Messenger messenger = new Messenger(new HttProtocol(new RequestParser(), new Router(), new ResponseFormatter()));
```
mrfishball, line 18: Nitpick, consider break these into separate lines:
```java
HttpProtocol protocol = new HttProtocol(
  new RequestParser(),
  new Router(),
  new ResponseFormatter()
);

Messenger messenger = new Messenger(protocol);
```

app/src/main/java/waiter/Router.java
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/17/commits/6fd747773a9adb7e5d4ef5e1a21cfb63ff761dcb)

app/src/main/java/waiter/Router.java
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/17/commits/6fd747773a9adb7e5d4ef5e1a21cfb63ff761dcb)

app/src/main/java/waiter/App.java
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/17/commits/ea062f7676d20cddf0fc85eb10395b89a9365b88)

.circleci/config.yml
```diff
@@ -26,7 +26,7 @@ jobs:
             gem install bundler
             bundle install --path vendor/bundle
             sudo apt install lsof
-            bundle exec spinach --tags simple-get
+            bundle exec spinach --tags simple-get, head-to-simple-get
```
mrfishball, line 29: I don't think the tag `head-to-simple-get` exists, might just be `simple-head`. https://github.com/8thlight/http_server_spec/blob/master/features/01_getting_started/simple_head.feature

app/src/test/java/waiter/RouterTest.java
```diff
@@ -29,4 +29,16 @@ void simpleGetWithBodyRoute() {
         assertEquals("Content-Length: 11", route.getHeaders());
         assertEquals("Hello world", route.getBody());
     }
+
+    @Test
+    void headSimpleGetReturnsRouteWithoutBody() {
```
mrfishball, line 34: Should we add a test for no URL found?

jonmanahan: Sure! Ill add a test for both 500 and 404

.circleci/config.yml
mrfishball: Make sense, no worries!

app/src/test/java/waiter/RouterTest.java
```diff
@@ -29,4 +29,40 @@ void simpleGetWithBodyRoute() {
...
+        Route route = new Router().getRoute(request);
+
+        assertEquals("HTTP/1.1", route.getProtocol());
+        assertEquals("500 Internal Server Error", route.getStatus());
+        assertEquals("Content-Length: 0", route.getHeaders());
+        assertEquals("", route.getBody());
+    }
+
+    @Test
+    void NoUrlFoundReturnsNotFoundRouteWithoutBody() {
```
mrfishball, line 58: Nice! Thank you!

app/src/test/java/waiter/RouterTest.java
```diff
@@ -29,4 +29,40 @@ void simpleGetWithBodyRoute() {
...
+        Route route = new Router().getRoute(request);
+
+        assertEquals("HTTP/1.1", route.getProtocol());
+        assertEquals("200 OK", route.getStatus());
+        assertEquals("Content-Length: 0", route.getHeaders());
+        assertEquals("", route.getBody());
+    }
+
+    @Test
+    void NoMethodOrUrlFoundReturnsInternalServerErrorRouteWithoutBody() {
```
mrfishball, line 46: I'm not sure this is quite right. If the route is not found, then it should be 404. If the route exists, but the method is not supported then we should return a [405](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405) instead.

jonmanahan: That is true, but I thought this was testing if both the route and the method are not supported

mrfishball: In that case, if a route is not found, regardless of method, we should always return 404 instead of a 500

mrfishball: For example, you can try in Postman, doing a `POST` to `google.com/bar` and it'll return a 404.

mrfishball (approved):
Other than ensuring we always return 404 for not found, the rest looks good! Thanks Jon!

app/src/test/java/waiter/RouterTest.java
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/17/commits/bf3be9d6f0dac707cde56cd7b18a4485db55b49b)

jonmanahan:
> Other than ensuring we always return 404 for not found, the rest looks good! Thanks Jon!

Should be all good to go now! Thank you for all of the help and the review, very much appreciated! 😄

PR #18, opened 2022-06-17
Builder pattern refactor

jonmanahan:


app/src/main/java/waiter/RouteDirector.java
```diff
@@ -0,0 +1,54 @@
...
+                .build();
+    }
+
+    public Route constructHeadSimpleGet(RouteBuilder routeBuilder) {
+        return routeBuilder.newUp()
+                .url("/simple_get")
+                .method("HEAD")
+                .status("200 OK")
+                .headers("Content-Length:")
+                .body("")
```
elijahsullivan, line 21: It's interesting to me that I need to define an empty body for this `HEAD` method.

If I were to specify a value in this body and make a `HEAD` request to `simple_get` would the response include a body?

app/src/main/java/waiter/RouteDirector.java
```diff
@@ -0,0 +1,54 @@
...
+                .build();
+    }
+
+    public Route constructUrlExistsButNoMethodFound(RouteBuilder routeBuilder, String url) {
+        return routeBuilder.newUp()
+                .url(url)
+                .method("")
+                .status("404 Not Found")
+                .headers("Content-Length:")
```
elijahsullivan, line 50: What happens if I don't specify this `Content-Length` header when creating a new route? Does this `Content-Length:` header need to be specified for every route? If so, how might we be able to remove this duplication?

app/src/main/java/waiter/RouteDirector.java
```diff
@@ -0,0 +1,54 @@
+package waiter;
+
+public class RouteDirector {
+
+    public Route constructGetSimpleGet(RouteBuilder routeBuilder) {
+        return routeBuilder.newUp()
+                .url("/simple_get")
+                .method("GET")
+                .status("200 OK")
```
elijahsullivan, line 9: Whenever I create a new route to a resource with a `GET` or `HEAD` method will I need to remember to add a "200 OK" status code?

app/src/main/java/waiter/RouteDirector.java
jonmanahan: That is another really good point, as of right now if `Content-Length` is not specified we get that `no chunk, no close, no size. Assume close to signal end` section in the response that we were getting before. So I would say in order to get a proper response, it would want to be specified. As a result I was thinking of looking into making it a default configuration for any new route!

app/src/main/java/waiter/RouteDirector.java
jonmanahan: As of right now this is the case, the status code and message need to be specified for every route

mrfishball: Should we use Enum for this? Usage can be: `StatusCode.OK` or `StatusCode.NotFound` etc.

app/src/main/java/waiter/RouteDirector.java
mrfishball: I'm curious, if I have a route say `/foo` and it supports `GET`, `HEAD` and `POST`, does it mean I have to create 3 routes with each methods for the same resource?

app/src/test/java/waiter/RouterTest.java
```diff
@@ -10,59 +10,59 @@ public class RouterTest {
...
-        assertEquals("Content-Length: 0", route.getHeaders());
-        assertEquals("", route.getBody());
+        assertEquals("HTTP/1.1", Response.getProtocol());
+        assertEquals("200 OK", Response.getStatus());
+        assertEquals("Content-Length: 0", Response.getHeaders());
+        assertEquals("", Response.getBody());
     }
 
     @Test
     void NoMethodOrUrlFoundReturnsInternalServerErrorRouteWithoutBody() {
```
mrfishball, line 46: I don't think this description is accurate since we're returning 404 instead of internal server error.

app/src/test/java/waiter/RouterTest.java
```diff
@@ -10,59 +10,59 @@ public class RouterTest {
...
+        Response Response = new Router().getRequestedRoute(request);
 
-        assertEquals("HTTP/1.1", route.getProtocol());
-        assertEquals("404 Not Found", route.getStatus());
-        assertEquals("Content-Length: 0", route.getHeaders());
-        assertEquals("", route.getBody());
+        assertEquals("HTTP/1.1", Response.getProtocol());
+        assertEquals("404 Not Found", Response.getStatus());
+        assertEquals("Content-Length: 0", Response.getHeaders());
+        assertEquals("", Response.getBody());
```
mrfishball, line 66: What happens if you hit the resource in Postman or in the browser, would it just give you a blank page? I wonder if we should return a body that tells the user what happened.

app/src/main/java/waiter/RouteDirector.java
```diff
@@ -0,0 +1,54 @@
...
+                .headers("Content-Length:")
+                .body("Hello world")
+                .build();
+    }
+
+    public Route constructNoUrl(RouteBuilder routeBuilder) {
+        return routeBuilder.newUp()
+                .url("")
```
mrfishball, line 37: When and how will this route be used?

app/src/main/java/waiter/RouteDirector.java
mrfishball: To Eli's point, we probably want to be able to set the content length dynamically. The current implementation suggests that `content length` and `body` can be set independent of each other which is not quite right imo.

app/src/main/java/waiter/RouteDirector.java
```diff
@@ -0,0 +1,54 @@
...
+                .status("404 Not Found")
+                .headers("Content-Length:")
+                .body("")
+                .build();
+    }
+
+    public Route constructUrlExistsButNoMethodFound(RouteBuilder routeBuilder, String url) {
+        return routeBuilder.newUp()
+                .url(url)
+                .method("")
```
mrfishball, line 48: Same thing here, do we need this route?

app/src/main/java/waiter/Routes.java
```diff
@@ -0,0 +1,42 @@
...
+        }
+    }
+
+    public Routes constructRoutes() {
+        RouteDirector routeDirector = new RouteDirector();
+        RouteBuilder routeBuilder = new RouteBuilder();
+
+        this.addRoute(routeDirector.constructGetSimpleGet(routeBuilder));
+        this.addRoute(routeDirector.constructHeadSimpleGet(routeBuilder));
+        this.addRoute(routeDirector.constructGetSimpleGetWithBody(routeBuilder));
```
mrfishball, line 27: If I want to add a route say `/foo` do I have to update this class?

app/src/main/java/waiter/RouteDirector.java
mrfishball: What about content type? I don't see any reference for that at the moment. Is it something we need to worry about?

app/src/main/java/waiter/RouteDirector.java
jonmanahan: We need this route to cover our bases for if someone makes a call to the right url with the wrong or no method, without this the server would break if someone did something like a head request to simple_get_with_body.

app/src/test/java/waiter/RouterTest.java
jonmanahan: It would indeed just give a blank page, I could return a body that says something along the lines of `Resource could not be found, please try again`

app/src/main/java/waiter/Routes.java
jonmanahan: You would have to add an `addRoute` call to add the route `/foo`

app/src/test/java/waiter/RouterTest.java
jonmanahan: Good catch, changing this now, thank you! 😄

app/src/main/java/waiter/RouteDirector.java
mrfishball: That seems strange to me that we need to create a route for a route that doesn't exist. Is there a way we can get the same result without doing this?

app/src/main/java/waiter/RouteDirector.java
jonmanahan: I can look into this, I like the Enum suggestion, thank you! 😄

app/src/main/java/waiter/RouteDirector.java
mrfishball: I see. This also feels strange to me for the same reasons I gave in the other comment.

app/src/main/java/waiter/RouteDirector.java
jonmanahan: As of right now, I don't think the absence or presence of content type is something we need to worry about

app/src/main/java/waiter/RouteDirector.java
jonmanahan: As of right now, you would have to create 3 routes with each methods for the same resource

We do set the content length dynamically, but that is true, it does suggest that they can be set independently, I will try to fix this so that the content-length header is not specified at all.

app/src/main/java/waiter/RouteDirector.java
jonmanahan: Hmmm, while I agree this could be confusing, I am confused as to how the server will know how to respond to a resource that doesn't exist if I don't have a route for it?

app/src/test/java/waiter/RouterTest.java
mrfishball: @matteeyao @elijahsullivan what do you all think?

app/src/main/java/waiter/RouteDirector.java
elijahsullivan: +1 to the last point made by @mrfishball

I'll add that this feels to me like we're building the response when routes are being defined rather than when a request is received, parsed, etc.

app/src/test/java/waiter/RouterTest.java
elijahsullivan: @mrfishball Agreed

elijahsullivan: Here's the default 404 response from Next.js as an example.

app/src/main/java/waiter/RouteDirector.java
mrfishball: +1 to Eli's last point.

app/src/main/java/waiter/RouteDirector.java
mrfishball: This is unrelated to this PR, but if we're returning a body, we should also specify the content type per the [HTTP specifications](https://www.rfc-editor.org/rfc/rfc7231#section-3.1.1.5). Given the number of comments already there and to avoid expanding the scope of this PR: I'm fine if we address this in a separate PR. Thoughts?

app/src/main/java/waiter/Routes.java
jonmanahan: We now add routes at the top level (for the user), and we also no longer need to add a test for every route, addressed in [this commit](https://github.com/jonmanahan/waiter/pull/18/commits/3bc8754c8fee547caf992d65eb1c11005cecf743)

app/src/main/java/waiter/RouteDirector.java
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/18/commits/3bc8754c8fee547caf992d65eb1c11005cecf743)

app/src/main/java/waiter/RouteDirector.java
jonmanahan: Working on status enums now!

app/src/main/java/waiter/RouteDirector.java
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/18/commits/c961b693c2652acfc53b02518629f441292d8b50)

app/src/main/java/waiter/App.java
```diff
@@ -9,20 +9,54 @@
...
+                        .url("/simple_get")
+                        .methods(new String[]{"GET", "HEAD"})
+                        .handler(okHandler)
+                        .build()
+        );
+
+        routes.addRoute(
+                routeBuilder.newUp()
+                        .url("/simple_get_with_body")
+                        .methods(new String[]{"GET"})
```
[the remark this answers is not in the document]
jonmanahan: I think I used the Enums in [this commit](https://github.com/jonmanahan/waiter/pull/18/commits/c961b693c2652acfc53b02518629f441292d8b50), but I don't think Im using them to the best of their abilities since I am immediately converting them to strings, Ill look into making the utilize Enums throughout!

app/src/test/java/waiter/RouterTest.java
jonmanahan: Not nearly as fancy (I can look into this if desired), but addressed in [this commit](https://github.com/jonmanahan/waiter/pull/18/commits/9c79171c8dc6dfaefd49fa6020893886b000ccad)

app/src/main/java/waiter/RouteBuilder.java
```diff
@@ -20,37 +17,17 @@ public RouteBuilder url(String url) {
...
-    public RouteBuilder method(String method) {
-        this.method = method;
+    public RouteBuilder methods(String[] methods) {
+        this.methods = methods;
         return this;
     }
 
-    public RouteBuilder status(String status) {
-        this.status = status;
+    public RouteBuilder handler(Callable<Response> handler) {
```
mrfishball, line 25: Do we expect route without a method and/or handler?

jonmanahan: Im not sure I understand, apologies 😦, what do you mean? Thank you for all the help as well btw! 😄

app/src/main/java/waiter/Routes.java
```diff
@@ -2,41 +2,38 @@
...
-        this.addRoute(routeDirector.constructNoUrl(routeBuilder));
+    public boolean exists(Request request) {
+        return this.routes.containsKey(request.getUrl());
+    }
 
-        for (String url : this.routes.keySet()) {
-            this.addRoute(routeDirector.constructUrlExistsButNoMethodFound(routeBuilder, url));
+    public Response handle(Request request) {
+        Route route = this.routes.get(request.getUrl());
+        if(!route.methodExistsForUrl(request.getMethod())) {
```
mrfishball, line 25: Consider moving this to `Router` since it's already handling the route not found scenario. Having both `Router` and `Route` handle requests don't feel right. IMO `Router` should decide how each route is handled, `Routes` is a collection of `Route` and so it knows about all the routes there are and can provide information to the `Router` to help make the decision.

app/src/main/java/waiter/Routes.java
```diff
@@ -2,41 +2,38 @@
...
+            return new ResponseBuilder()
+                    .newUp()
+                    .status("404 Not Found")
```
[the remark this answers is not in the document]
jonmanahan: Only for now, basically a placeholder for 405 until after this PR! 😄

app/src/main/java/waiter/Routes.java
jonmanahan: Makes sense to me! Thank you 😄

app/src/main/java/waiter/RouteBuilder.java
mrfishball: For example, can I make a route without a handler:
```java
RouteBuilder
  .newUp()
  .methods(...)
  .build()
```

app/src/main/java/waiter/Routes.java
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/18/commits/509da0f689e34e8dd6bf4b266f5885afb275fde5)

mrfishball: 👍 Btw, feel free to spin up follow-up tickets to address concerns/comments that aren't immediately relevant to this PR. I definitely noticed the number of comments and don't want to make this be the "make the code perfect" PR.

app/src/main/java/waiter/RouteBuilder.java
jonmanahan: That is a good point, I can and i don't think I should be able to, I feel like url, methods, and handler should all be required, Im gonna look into making them required (if that makes sense), incredible catch, thank you a ton Steven! 😄

app/src/main/java/waiter/Routes.java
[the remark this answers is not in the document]
mrfishball: Thanks for clarifying!

app/src/main/java/waiter/App.java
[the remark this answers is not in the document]
jonmanahan: Im gonna look to separate PR this if thats all good with you, if not no worries, just figured this was on the lower end of priority, either way let me know 😄

app/src/test/java/waiter/ResponseBuilderTest.java
```diff
@@ -10,20 +10,26 @@
...
 
-public class ResponseFormatterTest {
+public class ResponseBuilderTest {
 
     @Property
     void responseProperlyFormatted(@ForAll @Size(4) List<@AlphaChars @NotBlank String> requestResults) {
-        Response Response = new Response(requestResults.get(0), requestResults.get(1), requestResults.get(2), requestResults.get(3));
+        Response response = new ResponseBuilder()
+                .newUp()
+                .protocol(requestResults.get(0))
```
mrfishball, line 19: What do `requestResults` and `requestResults.get(0)` look like?

jonmanahan: They are just random strings acquired via jqwiks property-based-testing, just to show that it doesn't matter what is provided, it will be formatted correctly if that makes sense

app/src/test/java/waiter/HttProtocolTest.java
```diff
@@ -1,78 +1,122 @@
...
-
-        assertTrue(response.contains("HTTP/1.1 200 OK"));
-        assertTrue(response.endsWith("\r\n\r\n"));
-    }
-
-    @Test
-    void getSimpleGetWithBodyReturnsOKWithBody() {
-        String fromClient = "GET /simple_get_with_body HTTP/1.1";
+    @Property
+    void returnsOkWithNoBodyWhenUrlAndMethodExist(@ForAll @Size(3) List<@AlphaChars @NotBlank String> requestFields) {
```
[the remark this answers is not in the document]
jonmanahan: Sorry if there is confusion with all of the commits :(

app/src/main/java/waiter/RouteBuilder.java
mrfishball: Yea I think maybe a route builder might not make sense in this case. Because We always expect `url`, `method` and `handler` to be there so maybe a normal function/method would do:
```java
routes.addRoute(url: ..., method: ..., handler: ...) {}
```

app/src/test/java/waiter/ResponseBuilderTest.java
mrfishball: Interesting. I'm curious if status excepts a enum type, would property testing still apply here?

mrfishball: Right. I think we might want to think about property tests as a supplement to the unit tests so they should be separated IMO. For this particular test, if we focus on the unit test aspect of it. It's not clear to me what those arguments are.

As for mocking enum, we don't want to mock `Enum` for the same reason we don't mock `Array` or `Map`.

app/src/main/java/waiter/App.java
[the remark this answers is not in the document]
mrfishball: The idea of using Enum is that status code and method, even though they are represented as strings to us, have specific formats and values. For example `200 OK` instead of `20 ok`, `GET` instead of `gET` or `get`. By using string we're saying, as long as they are string, they'll work as described in the HTTP specifications but that's not the case. Enum combined with Java's type safety helps us define that format and value so that there's no confusion as to what those requirements are. Also imagine if the specification changes, we only need to update one spot.

The purpose of Enum is essentially to make devs' life easier, the same with the builder's pattern. We could bypass all of that and construct a response in a string but that'll be tedious and error-prone.

app/src/test/java/waiter/ResponseBuilderTest.java
mrfishball: Yea I can pair for a bit today. My calendar should be updated.

app/src/main/java/waiter/RouteBuilder.java
jonmanahan: Remove routeBuilder addressed in [this commit](https://github.com/jonmanahan/waiter/pull/18/commits/048ea1c654dd2bf3e04dc7d587e08ab8c89648be)

app/src/test/java/waiter/ResponseBuilderTest.java
jonmanahan: Make tests more clear and have them utilize the Enums, addressed in [this commit](https://github.com/jonmanahan/waiter/pull/18/commits/f82da021d65b0d3677ac9447063b627b01c5b6db)

mrfishball (approved):
Thanks for being patient with all the comments! I think I've been a very picky reviewer lately 😄

app/src/main/java/waiter/Router.java
```diff
@@ -1,55 +1,26 @@
...
-
-    public Route getRoute(Request request) {
-
-        Map<String, Map<String, String>> methodAssociatedUrls = routes.getOrDefault(request.getMethod(), routes.get(noMethod));
-        Map<String, String> requestResult = methodAssociatedUrls.getOrDefault(request.getUrl(), methodAssociatedUrls.get(noUrl));
-
-        return new Route(request.getProtocol(), requestResult.get("status"), requestResult.get("headers"), requestResult.get("body"));
+public record Router(Routes routes) {
+
+    public Response getRequestedResponse(Request request) {
```
mrfishball, line 5: 👍 👍

jonmanahan:
> Thanks for being patient with all the comments! I think I've been a very picky reviewer lately 😄

Of course Steven, I appreciate all of the comments, improvements and help! Regardless of pickiness, you have been an immense help and alot of the important catches in this PR would have potentially gone through without you, I appreciate it all no matter what! Thank you 😄

jonmanahan:
Just a note here, found a bug in my code for if a HEAD request is made with the wrong url, it would return a response with a body (which can't be done for a HEAD) so I fixed that in [this commit](https://github.com/jonmanahan/waiter/pull/18/commits/d01dd44c3840e4878a707c8e8939f02c284ed8d6)

PR #22, opened 2022-06-29
Simple post

jonmanahan:


app/src/main/java/waiter/Connectable/ClientConnection.java
```diff
@@ -38,8 +40,36 @@ public void close() throws IOException {
         this.socket.close();
     }
 
-    private boolean notEndOfLine(StringBuilder requestStartLineBuilder) {
-        return requestStartLineBuilder.indexOf("\r\n") == -1;
+    private boolean notEndOfRequest(StringBuilder requestMessageBuilder) {
```
elijahsullivan, line 43: Non-blocking: Some of these private functions (here and below) seem like they could live in a class that is more specifically related to parsing requests rather than in the `ClientConnection`.

app/src/main/java/waiter/Request.java
```diff
@@ -27,4 +27,12 @@ public String getUrl() {
...
-}
+
+    public String getHeaders() {
+        return headers;
+    }
+
+    public String getBody() {
+        return body;
+    }
+}
```
elijahsullivan, line 38: Nitpick: Newline here.

app/src/test/java/waiter/HttProtocolTest.java
```diff
@@ -57,15 +67,50 @@ void returnsOkWithNoBodyWhenHeadMethodWithExistingUrl() {
...
+    @Test
+    void returnsOkWithBodyWhenPostMethodWithExistingUrl() {
+        String url = "/echo_body";
+        String body = "foo";
+        String request = String.format(
+                """
+                POST %s HTTP/1.1
+                Content-Length: %d
+                
+                %s""", url, body.length(), body).replace("\n", "\r\n");
```
elijahsullivan, line 79: Non-blocking: Not asking you to change this, but FWIW I find this somewhat harder to read than just the plain text of the request, i.e. no separate url, body, call to `replace`, etc.

app/src/main/java/waiter/Connectable/ClientConnection.java
mrfishball: +1

app/src/main/java/waiter/Connectable/ClientConnection.java
```diff
@@ -38,8 +40,36 @@ public void close() throws IOException {
...
+
+        return contentLength;
+    }
+
+    private boolean contentLengthExists(String[] contentLengthHeader) {
+        return contentLengthHeader.length == 2;
+    }
+
+    private boolean headersWereRead(StringBuilder requestMessageBuilder) {
+        return requestMessageBuilder.indexOf("\r\n\r\n") != -1;
```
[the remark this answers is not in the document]
mrfishball: Could you use `System.lineSeparator()` for these?

app/src/main/java/waiter/App.java
```diff
@@ -36,23 +36,33 @@ private static Routes constructRoutes() {
...
+                new Route("/simple_get_with_body", new Request.Method[]{Request.Method.GET}, okWithOnlyResponseBodyHandler)
         );
 
         routes.addRoute(
                 new Route("/head_request", new Request.Method[]{Request.Method.HEAD, Request.Method.OPTIONS}, okHandler)
         );
 
+        routes.addRoute(
+                new Route("/echo_body", new Request.Method[]{Request.Method.POST}, okWithRequestBodyHandler)
+        );
```
elijahsullivan, line 48: Non-blocking: I think your design will need to change here for a future use case where the handler for two methods (for the same endpoint) need to be different.

Example: Consider an endpoint with POST and PUT methods. Suppose that the handler for the POST method should write data to some datastore and that the handler for the PUT method should update some data in a datastore. You _could_ handle the case I just described by bifurcating on the methods within your handler function, but I would urge you not to do that.

app/src/test/java/waiter/HttProtocolTest.java
mrfishball: +1

app/src/test/java/waiter/HttProtocolTest.java
```diff
@@ -37,14 +42,19 @@ void returnsOkWithNoBodyWhenGetMethodWithExistingUrl() {
     @Test
     void returnsOkWithNoBodyWhenHeadMethodWithExistingUrl() {
```
mrfishball, line 43: How are these tests any different? What are we testing here?

mrfishball (approved):
Sorry about the late review!

app/src/main/java/waiter/Connectable/ClientConnection.java
```diff
@@ -38,8 +40,36 @@ public void close() throws IOException {
...
+            contentLength = Integer.parseInt(
+                    splitRequestMessage[1].substring(0, splitRequestMessage[1].indexOf("\r\n"))
+            );
+        }
+
+        return contentLength;
+    }
+
+    private boolean contentLengthExists(String[] contentLengthHeader) {
+        return contentLengthHeader.length == 2;
```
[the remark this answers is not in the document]
jonmanahan: This is checking for a 2 because I use a `String.split()` to find the "Content-Length" header, which I then also use the result of here to check whether or not it split properly properly (if the length is two it found the Content-length or else it would be one). Looking at it more though, the name for the argument I pass in is a little misleading and the way it is checked for here is inconsistent with the other "request portion exists call". So I think Im going to check for the "Content-Length" using `indexOf` like the other one and combine the two function to be a generic `existsInRequestString` or something if that makes sense.

app/src/main/java/waiter/Connectable/ClientConnection.java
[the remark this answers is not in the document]
jonmanahan: I wasn't aware this was a thing, thank you! Ill be sure to use it and also create a shared constant for them too. My thought is that hese will live in the RequestParser, along with some of the functions (talked about below) will move to too

app/src/main/java/waiter/Connectable/ClientConnection.java
jonmanahan: Now that you mention it I completely see it and agree, I was so focused on trying to make these named properly and single responsibility when it came to the individual functions that I missed they belong elsewhere, Ill be sure to move them over, thank you for the catch! 😄

jonmanahan:
> Sorry about the late review!

No worries at all, I appreciate the reviews none the less! Thank you guys 😄

app/src/test/java/waiter/HttProtocolTest.java
jonmanahan: These tests are each testing the response result to a method being used (one each for post, get, head, etc)

mrfishball: Wouldn't we want to test that as part of the `Router`? This doesn't seem like it should be the responsibility of the `HttpProtocol`.

app/src/main/java/waiter/Connectable/ClientConnection.java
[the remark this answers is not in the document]
jonmanahan: Note: I tried using `System.lineSeparator()` but it returns only `\n` for Linux, so I will be specifying the shared constant to equal "\r\n" and "\r\n\r\n" instead

app/src/test/java/waiter/HttProtocolTest.java
jonmanahan: Would we be able to pair on this/discuss this on zoom at some point (no rush at all), I have some questions/clarifications I think would be easier to talk about over chat 😄

app/src/main/java/waiter/Connectable/ClientConnection.java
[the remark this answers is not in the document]
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/24/commits/79efe62940ad06a107536e3f395856ecf8bdf2ab)

app/src/main/java/waiter/Request.java
jonmanahan: Addressed in [this commit](https://github.com/jonmanahan/waiter/pull/24/commits/0294239e9ea74aa0369ba050ca9fa0ec8851d0fd)

PR #23, opened 2022-07-01
Simple redirect

jonmanahan:


app/src/test/java/waiter/RouterTest.java
```diff
@@ -92,4 +92,28 @@ void returnsMethodNotAllowedWithReasonResponseForExistingUrlButNoGetMethodInRout
         assertEquals("Content-Length: 0", contentLengthHeader);
         assertEquals("", Response.getBody());
     }
+
+    @Test
+    void returnsMovedPermanentlyWithLocationHeaderForResourceThatHasBeenAssignedNewUrl() {
```
mrfishball, line 97: I wish we can test the `Router` without having to test every method permutations. For example, routing GET should imply that routing POST will work too.

mrfishball (approved):

app/src/test/java/waiter/RouterTest.java
jonmanahan: Hmmm, Ill have to think about this, my initial thought is to test based on the Method Enum if that makes sense, but Im not too sure

